PR #366, opened 2023-04-23
✨ Allow to set additional fsspec kwargs for cloud instances

Koncopd:
subj

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -61,4 +61,4 @@ repos:
     hooks:
       - id: pydocstyle
         args: # google style + __init__, see http://www.pydocstyle.org/en/stable/error_codes.html
-          - --ignore=D100,D101,D102,D103,D106,D107,D203,D204,D213,D215,D400,D401,D403,D404,D406,D407,D408,D409,D413
+          - --ignore=D100,D101,D102,D103,D106,D107,D203,D204,D213,D215,D400,D401,D403,D404,D406,D407,D408,D409,D412,D413
--- a/docs/guide/04-set-storage.ipynb
+++ b/docs/guide/04-set-storage.ipynb
@@ -127,6 +127,19 @@
    ]
   },
   {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7bd9997b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# root.fs contains the underlying fsspec filesystem\n",
+    "assert (\n",
+    "    ln.setup.settings.storage.root.fs.cache_regions  # set by lamindb to True for s3 by default\n",
+    ")"
+   ]
+  },
+  {
    "cell_type": "markdown",
    "id": "ae953b6a",
    "metadata": {},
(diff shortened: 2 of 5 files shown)


Koncopd:
It looks like there are problems wiht CI instances related to migration. On some instances lnschema-core is new and on some it is old, this is why CI fails.

falexwolf:
You can resolve this in two ways:
1. Either delete the CI instance
2. Or migrate the CI instance

falexwolf:
Always feel free to delete the CI instances!

Another important thing is that the CI runs run on the latest version of lamindb. Clearing the cache often helps with that.

falexwolf:
I hope we'll get this under better control soon.

Koncopd:
Ok, understood.

Koncopd:
@falexwolf this is rather a quick hack.
```
lndb.setup.storage("s3://some-bucket")
lndb.setup.settings.storage.set_fs_kwargs(profile="some_profile", cache_regions=True) # any fsspec args
```
also can be used after `init` and `load` of postgres intstances.

It is quite complicated to add this on init, save it etc (especially in CLI).

This should do the trick for now.

falexwolf:
I like this!

falexwolf:
There is no need at all to do this during `init`! I prefer that it's called after, the way you did it!

falexwolf:
There is also no need to call it on the CLI. This is a highly configured case, we can ask users to use the Python API.

falexwolf:
We just need to add this documentation within the `set-storage` notebook, so that users can see it.

falexwolf:
Ohhhhh!

falexwolf:
I thought you went to bed! I just did a small refactor. Can you check?

falexwolf:
The refactor is just meant to be small simplification on the user end that allows them to pass it directly to `ln.setup.set.storage(root, **fs_kwargs)` rather than having to call another function.

Koncopd:
Makes making a test a bit more complicated though.

falexwolf:
Why not just adding the kwargs to `add_storage` in the test? Is that a problem?

falexwolf:
Ah, damn it, made the function private in the last minute. Will you fix or shall I? Don't want to cause conflicts.

Koncopd:
I did, but we need to check the kwarg also **before** setting.

Koncopd:
I will change.

Koncopd:
The problem with with approach is that sometimes setting some additional fsspec kwargs could be also needed for sqlite cloud instances on `init`.

falexwolf:
Then shall we just make `set_fs_kwargs` public again? And Thomas can just use the simplified way of passing to `set.storage` and the use case that you're alluding to can work through `Storage.set_fs_kwargs`?

falexwolf:
I'm not so worried about sqlite cloud instances right now, though. Because nobody asked for it (of course it's good that you think for the future and want to minimize technical debt!).

Koncopd:
`set_fs_kwargs`  won't help sqlite instances. On `init` they will need all kwargs set already to have proper access rights etc.

Koncopd:
Can be merged, i guess.

falexwolf:
Great! Then let's merge!

PR #378, opened 2023-05-21
✨ Add track command to CLI

Koncopd:
`lndb track nb.ipynb --pypackage lamindb,bionty --parent parent_id`
The command initializes nbproject metadata in `nb.ipynb` to avoid writing it in `ln.track()`

It doesn't create `Transform` or `Run`, i think it is better to do from the notebook itself.

### DIFF
--- a/lndb/__main__.py
+++ b/lndb/__main__.py
@@ -91,12 +91,8 @@ aa("action", choices=["generate"], help="Generate migration.")
 # track anotebook (init nbproject metadata)
 track_parser = subparsers.add_parser("track", help=track_help)
 aa = track_parser.add_argument
-nb_path_help = "A path to the notebook to track."
-aa("nb_path", type=str, metavar="filepath", help=nb_path_help)
-parent_help = (
-    "One or more (delimited by ',') ids of direct ancestors in a notebook pipeline"
-)
-aa("--parent", type=str, metavar="parent", default=None, help=parent_help)
+filepath_help = "A path to the notebook to track."
+aa("filepath", type=str, metavar="filepath", help=filepath_help)
 pypackage_help = "One or more (delimited by ',') python packages to track."
 aa("--pypackage", type=str, metavar="pypackage", default=None, help=pypackage_help)
 
@@ -153,7 +149,7 @@ def main():
         if args.action == "generate":
             return migrate.generate()
     elif args.command == "track":
-        track(args.nb_path, args.parent, args.pypackage)
+        track(args.filepath, args.pypackage)
     else:
         logger.error("Invalid command. Try `lamin -h`.")
         return 1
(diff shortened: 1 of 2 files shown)


lndb/__main__.py
```diff
@@ -87,6 +88,18 @@
...
 
+# track anotebook (init nbproject metadata)
+track_parser = subparsers.add_parser("track", help=track_help)
+aa = track_parser.add_argument
+nb_path_help = "A path to the notebook to track."
+aa("nb_path", type=str, metavar="filepath", help=nb_path_help)
+parent_help = (
+    "One or more (delimited by ',') ids of direct ancestors in a notebook pipeline"
+)
+aa("--parent", type=str, metavar="parent", default=None, help=parent_help)
```
falexwolf, line 99: I'd not support the parent argument as this is something that shouldn't be needed in the context of lamindb!

lndb/_notebook.py
```diff
@@ -0,0 +1,26 @@
+from typing import Optional
+
+from lamin_logger import logger
+
+
+def track(
+    notebook_path: str, parent: Optional[str] = None, pypackage: Optional[str] = None
```
falexwolf, line 7: I'd remove the parent arg here!

lndb/_notebook.py
```diff
@@ -0,0 +1,26 @@
...
+):
+    try:
```
falexwolf, line 9: nbproject is a requirement of lamindb; hence, this isn't needed - but of course it doesn't hurt much! :)

lndb/_notebook.py
```diff
@@ -0,0 +1,26 @@
...
+        from nbproject.dev import initialize_metadata, read_notebook, write_notebook
+    except ModuleNotFoundError:
+        logger.warning("Install nbproject! pip install nbproject")
+        return None
+
+    nb = read_notebook(notebook_path)
+    if "nbproject" not in nb.metadata:
+        if parent is not None:
```
falexwolf, line 17: This logic can disappear if we ditch parent here.

lndb/__main__.py
```diff
@@ -87,6 +88,18 @@
 aa = migr.add_argument
 aa("action", choices=["generate"], help="Generate migration.")
 
+# track anotebook (init nbproject metadata)
+track_parser = subparsers.add_parser("track", help=track_help)
+aa = track_parser.add_argument
+nb_path_help = "A path to the notebook to track."
+aa("nb_path", type=str, metavar="filepath", help=nb_path_help)
```
falexwolf, line 95: Can we rename nb_path to filepath?

lndb/_notebook.py
Koncopd: Removed

lndb/_notebook.py
Koncopd: Yes, but not of lndb. Just to avoid any strange edge cases.

lndb/_notebook.py
Koncopd: Removed

lndb/__main__.py
Koncopd: Renamed

PR #401, opened 2023-06-07
⚡ Upload sqlite file in separate process

falexwolf:
- https://laminlabs.slack.com/archives/C04FPE8V01W/p1686136203501999

### DIFF
--- a/lamindb_setup/dev/_settings_instance.py
+++ b/lamindb_setup/dev/_settings_instance.py
@@ -26,6 +26,14 @@ from .upath import UPath
 #         cursor.close()
 
 
+def upload_and_sync_timestamp(local_file: Path, cloud_file: UPath):
+    cloud_file.upload_from(local_file)  # type: ignore
+    cloud_mtime = cloud_file.modified.timestamp()  # type: ignore
+    # this seems to work even if there is an open connection
+    # to the cache file
+    os.utime(local_file, times=(cloud_mtime, cloud_mtime))
+
+
 class InstanceSettings:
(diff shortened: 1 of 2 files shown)


lamindb_setup/dev/_settings_instance.py
```diff
@@ -98,17 +106,21 @@ def _sqlite_file_local(self) -> Path:
...
+            cache_file = self.storage.cloud_to_local_no_update(cloud_file)
+
+            if new_process:
+                from multiprocessing import Process
+
+                Process(
+                    target=upload_and_sync_timestamp, args=(cache_file, cloud_file)
+                ).start()
+            else:
+                upload_and_sync_timestamp(cache_file, cloud_file)
```
falexwolf, line 123: @Koncopd!

Koncopd:
There is a problem with the sync notebook, but i don't really understand why this happens now, the update of local sqlite file should not be affected.

Koncopd:
Now it is probably better not to run this notebook at all.

Koncopd:
image
Running locally it fails on Windows. I have no idea why it still work on linux / CI on main.

falexwolf:
I reproduce the error on CI locally on my mac depending on how fast I run the notebook 😆 - if I run it slowly and allow the upload process to finish everything is good. If I run it fast and the upload process hasn't yet finished and the time stamp is written, it evidently fails
image

This makes me think: OMG, this is dangerous without any await logic among processes.

And that makes me think: OMG, this is going to be complicated...

PR #402, opened 2023-06-07
🚸 Simplify remote SQLite synching & locking

falexwolf:
By having only 3 points in which we synch & lock/unlock the remote SQLite file, we can simplify the code dramatically and still create a safe & performant user experience.

The one and only downside is that the user needs to manually push changes back to the cloud by calling `lamin close`. But given this is a known journey (`git push`), it shouldn't be a big deal.

The 3 points are:
* `init` merely locks the remote sqlite file but doesn't yet upload it
* `load` synchs changes from a remote sqlite file to the local file & locks the remote file
* `close` updates the remote sqlite file & unlocks it

In this PR, we don't put forward the locking on init because it is heavily used in concurrent CI settings.

The really great thing here is that `lamindb` can be entirely free from logic around synching the cloud sqlite. It makes the code simpler and less error prone, and very performant.

### DIFF
--- a/docs/faq/test-sqlite-sync.ipynb
+++ b/docs/faq/test-sqlite-sync.ipynb
@@ -75,6 +75,82 @@
    ]
   },
   {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "45967bdd",
+   "metadata": {},
+   "source": [
+    "Remote SQLite file does not yet exist upon instance init:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "037ac4f5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "assert not settings.instance._sqlite_file.exists()"
+   ]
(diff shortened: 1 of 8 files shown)


falexwolf:
So, I'm pretty happy with the user story in `test-sqlite-sync` notebook.

From my perspective, the only remaining thing is to have a standard naming convention for "cloud SQLite".

Given GitHub uses "remote repository" and we already talk about "remote instances", it might be more natural to consistently call things remote_sqlite rather than cloud_sqlite.

The other thing is allowing multiple lndb files in a bucket, restricting the locker just to that single lndb file.

falexwolf:
This is the first time the `_exclusion` code actually gets tested, @Koncopd
image

https://app.codecov.io/github/laminlabs/lamindb-setup/commit/ddecc00d0a6e379d643d02030c6a270f1e4aef92

Could you think about how to get this to 100% coverage when reviewing this?

PR #403, opened 2023-06-07
♻️ Rename and set empty locker for now

falexwolf:


Koncopd:
There is an import from `_exclusion` in `__init__.py` also, maybe in other places too, so should be renamed there also.

falexwolf:
Nice, GitHub seems down here:
image

falexwolf:
Even github itself shows it correctly, but doesn't trigger the action here:
image

falexwolf:
That's crazy, I merged this one now, but it didn't take the last commit into account despite it being on the remote `amend` branch...

falexwolf:
Making a new PR: https://github.com/laminlabs/lamindb-setup/pull/404

PR #415, opened 2023-06-11
🔥 Adapt locker to lock entire lamindb session

Koncopd:
Now 
`settings.instance.cloud_sqlite_locker.lock()` doesn't lock anything by itself, it just check if it can obtain write access to the (cloud sqlite) instance.
After calling `settings.instance.cloud_sqlite_locker.lock()` at the load of an instance (internally), all write operations in lamindb should check the property `settings.instance.cloud_sqlite_locker.has_lock` before proceeding. 

upd:
Now prevents an instance from loading completely if locked by another user.

### DIFF
--- a/lamindb_setup/_load_instance.py
+++ b/lamindb_setup/_load_instance.py
@@ -81,6 +81,11 @@ def load(
             )
             return "instance-not-reachable"
 
+    # at this point the lock should be already initialized
+    if not isettings._cloud_sqlite_locker.has_lock:
+        locked_by = isettings._cloud_sqlite_locker._locked_by
+        raise RuntimeError(f"Can't load the instance, it is locked by {locked_by}.")
+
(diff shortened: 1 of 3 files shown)


Koncopd:
@falexwolf We have `unlock` in `_update_cloud_sqlite_file` but it is called not only by `close` but also by `setup_django`, so it is probably incorrect.

falexwolf:
It is correct!

falexwolf:
Does the comment I put not make sense? Happy to do a brief call!
https://github.com/laminlabs/lamindb-setup/blob/696e6f0bb853bf5de1280c6d69c617a77f5bd61a/lamindb_setup/dev/_django.py#L67-L68

falexwolf:
Deploy migration command means: `lamin migrate deploy`. This is an isolated admin command that

PR #421, opened 2023-06-15
🎨  Remove `lnhub_rest` from `lamindb_setup`

bpenteado:


falexwolf (approved):
This seems great!

Can you rerun with `lnhub_rest` from the dependencies in `pyproject.toml` removed?

falexwolf:
And, as a check, could you also see whether the CLI got slower through this? https://laminlabs.slack.com/archives/C04A0RMA0SC/p1685899943000489

bpenteado:
> This seems great!
> 
> Can you rerun with `lnhub_rest` from the dependencies in `pyproject.toml` removed?

Done!

pyproject.toml
```diff
@@ -19,7 +19,8 @@ dependencies = [
     "appdirs",
     "pandas",  # remove!!!
     "python-dateutil",
-    "universal_pathlib"
+    "universal_pathlib",
+    "postgrest"
```
falexwolf, line 23: I'm pretty sure you'll need the full supabase-py here, right now.

Strictly speaking, we'll only need gotrue and postgrest, but that's another refactor.

falexwolf: You'll need to copy these two lines: https://github.com/laminlabs/lnhub-rest/blob/2dba53b7696fdf17d1e422a7bcb57a68a02d7990/pyproject.toml#L15-L16

falexwolf:
You shouldn't need sqlmodel, but you should need pyjwt.

falexwolf:
:D https://github.com/laminlabs/lamindb-setup/pull/421#discussion_r1232333467

falexwolf:
> [The need stems from bionty tests, it seems](https://github.com/laminlabs/lamindb-setup/blob/5976ba82f65c6f8ce6078a0dbf7f003d4486b5cd/tests/test_bionty.py#L1)? I did not write these tests.

OMG, please remove this deadline!

@sunnyosun, can we add back bionty tests with Django now?

pyproject.toml
```diff
@@ -40,6 +40,7 @@ test = [
     "pytest-cov",
     "nbproject-test>=0.4.3",
     "sqlmodel",
```
falexwolf, line 42: Please remove sqlmodel and the whole dead code in the bionty test! It needs to be re-introduced based on Django - we somehow didn't notice this.

pyproject.toml
```diff
@@ -40,6 +40,7 @@ test = [
...
+    "pyjwt",
```
falexwolf, line 43: I think pyjwt needs to be a dependency not a test-dependency! Can you move it into the actual dependencies, please?

pyproject.toml
```diff
@@ -25,7 +19,8 @@ dependencies = [
     "appdirs",
     "pandas",  # remove!!!
     "python-dateutil",
-    "universal_pathlib"
+    "universal_pathlib",
+    "supabase",
```
falexwolf, line 23: Could you please pin supabase==1.0.3 as we did in the hub? We ran into things breaking because we hadn't pinned it at some point.

falexwolf:
After the small remaining changes, I think we can merge this! :D

It should speed up `lamin init` and `lamin load` considerably, according to the previous profiling I did.

And we now isolated all this client-side functionality and can refactor it correspondingly. I'll give this a pass once it's merged towards simplifying as much as possible.

falexwolf:
I tested profiling manually. It's still fast!

falexwolf:
This is good to merge from my end. Further refactors and simplifications can be done in subsequent PRs.

In the process of consolidating lamindb installation, this is a good step forward!

PR #440, opened 2023-07-17
👷 Run CI against the staging environment

bpenteado:


lamindb_setup/dev/_hub_client.py
```diff
@@ -89,9 +89,13 @@ def get_access_token(email: Optional[str] = None, password: Optional[str] = None
 
 
 def lamindb_client_config_settings(settings: BaseSettings) -> Dict[str, Any]:
-    connector_file, _ = urlretrieve(
-        "https://lamin-site-assets.s3.amazonaws.com/connector.env"
-    )
+    if os.environ["LAMIN_ENV"] == "staging":
```
falexwolf, line 92: This here is problematic as users won't have "LAMIN_ENV" set, and then this will error.

falexwolf: You have to use `os.getenv("LAMIN_ENV")` here, instead. This won't raise an error but return `None` when an env variable doesn't exist.

noxfile.py
```diff
@@ -1,24 +1,10 @@
...
-LAMIN_ENV = "prod"
-if "GITHUB_REF_NAME" in os.environ:
-    if os.environ["GITHUB_REF_NAME"] == "main":
-        LAMIN_ENV = "prod"
-    elif os.environ["GITHUB_REF_NAME"] == "staging":
-        LAMIN_ENV = "staging"
-elif "LAMIN_ENV" in os.environ:
-    LAMIN_ENV = os.environ["LAMIN_ENV"]
-
-env = {"LAMIN_ENV": LAMIN_ENV}
```
falexwolf, line 19: It's nice that all of this up there is gone!

PR #449, opened 2023-08-01
🎨 Fix `init-instance` anti-patterns and expand tests

bpenteado:


tests/test_init_instance.py
```diff
@@ -89,6 +89,19 @@ def test_init_instance_postgres_cloud_aws_europe():
     )
 
 
+def test_init_instance_sqlite():
```
falexwolf, line 92: @bpenteado, I fear this test wouldn't have captured the bug that Fred ran into.

The functionality that we wanted to cover is to register an SQLite instance on the hub, not to init a local SQLite instance.

While it doesn't hurt to also add a unit test, initializing a local SQLite instance is already covered [in the notebook tests](https://github.com/laminlabs/lamindb-setup/blob/main/docs/guide/01-init-instance.ipynb).

bpenteado: Fixed here: https://github.com/laminlabs/lamindb-setup/pull/450

PR #457, opened 2023-08-07
🐛 Fix aws access with no credentials

Koncopd:
subj


Koncopd:
Ok, this should solve the problems with the access to public s3 buckets.
This should be tested, but not sure how.

Probably we should also do something similar for gs.

PR #464, opened 2023-08-13
Enable direct password passing

Zethson:
Fixes https://github.com/laminlabs/lamindb-setup/issues/465

I personally prefer this. Any opinions?

Would need changes in https://github.com/laminlabs/laminci/blob/main/laminci/_nox.py and so on. Maybe I could make it more backwards compatible


Zethson:
It's not clear yet what the ultimate login experience will be so we're closing this for now

PR #467, opened 2023-08-13
🚸 Add delete confirmation dialogue

Zethson:
Fixes https://github.com/laminlabs/lamindb-setup/issues/295

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
         lamin_env:
           - "staging"
           - "prod"
-    timeout-minutes: 35
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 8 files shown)


Zethson:
> ImportError: cannot import name 'ParentsAware' from 'lnschema_core.models' (/opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/site-packages/lnschema_core/models.py)

Not sure where this is coming from now

falexwolf:
Thank you, @Zethson, this is great!

falexwolf:
I think tests fail because of an inconsistency between lnschema-core submodule and lnschema-bionty install. I think I should meanwhile remove the lnschema-core submodule or find another way that this can't happen anymore

falexwolf:
No, I see it's not that. Hmmm.

falexwolf:
Let's see if this some flakiness here - I haven't encountered any for a good while

Zethson:
I assume there's a lamin delete that expects an answer aka a "yes" somewhere that I missed? Currently afk

falexwolf:
Oh, right. That's very likely the reason. Would you search for it? This is otherwise good to merge from my point of view!

PR #479, opened 2023-08-17
👷 Run hub tests from `lamindb-setup`

bpenteado:


noxfile.py
```diff
@@ -26,15 +26,15 @@ def install(session: nox.Session, group: str) -> None:
             *"pip install --no-deps git+https://github.com/laminlabs/lnschema-core"
             .split()
         )
-        session.run(*"pip install .[aws,dev]".split())
+        session.run(*"pip install .[aws,dev,test]".split())
```
falexwolf, line 29: We don't need to globally update all test heavy hub test dependencies for all workers. It's enough to just install them for the hub worker!

PR #482, opened 2023-08-21
🐛 Fix UPath check on Storage init

Koncopd:
subj

### DIFF
--- a/tests/test_load_persistent_instance.py
+++ b/tests/test_load_persistent_instance.py
@@ -5,5 +5,5 @@ import lamindb_setup as ln_setup
 
 
 def test_load_persistent_instance():
-    ln_setup.load("laminapp-admin/persistent-instance-to-test-load")
+    ln_setup.load("testuser1/lamin-site-assets")
     ln_setup.close()
--- a/tests/test_load_persistent_instance.py
+++ b/tests/test_load_persistent_instance.py
@@ -1,5 +1,6 @@
-# this tries to load a persistent instance created with
-# lamin init --storage s3://lndb-setup-ci --name persistent-instance-to-test-load
+# this tries to load a persistent instance created in a public s3 bucket
+# with s3:GetObject and s3:ListBucket policies enabled for all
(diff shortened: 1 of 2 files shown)


Koncopd:
Had to add s3:ListBucket permission to lamin-site-assets for this to work.

tests/test_load_persistent_instance.py
```diff
@@ -1,5 +1,6 @@
-# this tries to load a persistent instance created with
-# lamin init --storage s3://lndb-setup-ci --name persistent-instance-to-test-load
+# this tries to load a persistent instance created in a public s3 bucket
+# with s3:GetObject and s3:ListBucket policies enabled for all
```
falexwolf, line 2: Thanks for this comment, Sergei!

falexwolf:
@Koncopd, are the 3 additional misses _real_? Would you maybe need to adapt a test a bit?

Our policy is to no longer merge any new un-covered code; we have way too much of it still and need to get to 100% coverage.

falexwolf:
Is it possible that the 3 lines were covered by accident before?
image

falexwolf:
Codecov reports no changes for these lines:
image

falexwolf:
Right, these are [all indirect changes](https://app.codecov.io/gh/laminlabs/lamindb-setup/pull/482/indirect-changes?src=pr&el=tree-more&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=laminlabs):
image

falexwolf:
One topic - which is on my plate - is that before the warning on missing migrations was raised, and not it is no longer (in the django file):
image

falexwolf:
Because the locker has so little coverage, it's hard to say where the other 2 lines come from: https://app.codecov.io/gh/laminlabs/lamindb-setup/pull/482/blob/lamindb_setup/dev/_cloud_sqlite_locker.py

Testing this properly is a whole work package, I guess, so, let's postpone this a bit.

PR #486, opened 2023-08-22
♻️ Introduce `create_path`

falexwolf:
- https://github.com/laminlabs/lamindb/pull/1072

This is likely not the ultimate solution, but should be good for a while.

It's hard to subclass UPath.

I think it's also overkill to create a global `Permissions` class at this point.

Essentially, it's as simple as checking whether AWS credentials are present, and that's happening now.

We could extend this to GCP & Azure if needed at some point.

### DIFF
--- a/lamindb_setup/_init_instance.py
+++ b/lamindb_setup/_init_instance.py
@@ -16,6 +16,7 @@ from .dev import InstanceSettings
 from .dev._docs import doc_args
 from .dev._settings_storage import StorageSettings
 from .dev._setup_schema import get_schema_module_name, load_schema
+from .dev.upath import create_path
 
 
 def register_storage(ssettings: StorageSettings):
@@ -240,11 +241,7 @@ def infer_instance_name(
         # better way of accessing the database name?
         return str(db).split("/")[-1]
 
-    if isinstance(storage, str):
-        storage_path = StorageSettings._str_to_path(storage)
-    else:
-        storage_path = storage
-
+    storage_path = create_path(storage)
     if isinstance(storage_path, UPath):
         name = storage_path._url.netloc
     else:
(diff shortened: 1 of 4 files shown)


Koncopd:
Could you avoid changing `_str_to_path`? This should not do any access check.

lamindb_setup/_init_instance.py
```diff
@@ -240,11 +241,7 @@ def infer_instance_name(
         # better way of accessing the database name?
         return str(db).split("/")[-1]
 
-    if isinstance(storage, str):
-        storage_path = StorageSettings._str_to_path(storage)
-    else:
-        storage_path = storage
-
+    storage_path = create_path(storage)
```
Koncopd, line 244: I think `create_path` is not needed here, only `str_to_path` is enough.

falexwolf: The way it's typed right now it also accepts UPath & Path. And str_to_path only accepts string, that's why I wrote it like this.

I agree that the additional checks aren't needed, but they also don't hurt because they're cached now.

I also agree that one could write this differently, I just wanted to eliminate the redundant if condition.

falexwolf:
This PR eliminates 5 lines in total, among which 1 uncovered line.

falexwolf:
That's not a great achievement, but I do think that the logic of "AWS credentials" is better handled in a global setting, agnostic to any StorageSettings object.

lamindb_setup/_init_instance.py
Koncopd: ok, great, agree

Koncopd:
So the `PermissionSettings` class is not needed with this, right?

falexwolf:
I think so, not anytime soon, I believe.

PR #492, opened 2023-08-29
🚸 Cache instance id during `load`

bpenteado:


bpenteado:
@falexwolf requesting your review here just to make sure this is in line with what we agreed on!

lamindb_setup/dev/_settings_instance.py
```diff
@@ -44,6 +46,11 @@ def __repr__(self):
                 representation += f"\n- {attr}: {value}"
         return representation
 
+    @property
+    def id(self) -> Union[None, str]:
```
falexwolf, line 50: I'd not expose it to the user for now, Bernardo. We might have a base62 id in the future, and then things will be confusing.

The instance UUID is a non-user facing internal thing.

falexwolf:
It looks good! Just, please add an assert statement somewhere in the tests. There are persistent instances used, e.g., `lamin-site-assets`.

PR #493, opened 2023-08-31
🚸 Cache instance id during init already

falexwolf:


lamindb_setup/dev/_hub_core.py
```diff
@@ -102,26 +102,26 @@ def init_instance(
...
         if instance is not None:
-            return "instance-exists-already"
+            return instance["id"]
 
         validate_unique_sqlite(
             hub=hub, db=db, storage_id=storage_id, name=name, account=account
         )
 
-        instance_id = uuid4().hex
+        instance_id = str(uuid4())
```
falexwolf, line 124: Keep `.hex`

tests/unit/test_init_instance.py
```diff
@@ -62,6 +62,7 @@ def test_init_instance_postgres_default_name(get_hub_client):
     assert instance["db_port"] == 5432
     assert instance["db_database"] == "pgtest"
     # client checks
+    assert ln_setup.settings.instance._id == instance["id"]
```
falexwolf, line 65: RHS needs to read `UUID(instance["id"]).hex`

PR #497, opened 2023-09-05
👔 Always let permission query from hub override local cache for remote instances

bpenteado:


falexwolf:
@bpenteado, this looks great & simple. Just: could you add a test? We probably need to spread it across 2 Jupyter notebooks because we'll first need to load it from the hub, then revoke permission in the hub, and then test that the instance can't be loaded anymore.

If there is a better way of testing it, we could of course do something else.

If it sounds like Voodoo I'm also happy to do this myself if you point me to the endpoint that I need to call to revoke permission!

PR #498, opened 2023-09-05
🔒 Require users to provide database credentials on first loading

fredericenard:
As discussed with @bpenteado it works well for admin users, but not for the others as RLS policies on account_instance table prevent them to update the db_user_id field. 

@falexwolf could you try locally and validate the UX?

I will add tests once RLS problems are solved and UX will be validated.

### DIFF
--- a/lamindb_setup/__main__.py
+++ b/lamindb_setup/__main__.py
@@ -11,6 +11,8 @@ init_help = "Init & config instance with db & storage."
 load_help = "Load instance by name."
 set_storage_help = "Set storage used by an instance."
 load_storage_help = "Load the instance with an updated default storage."
+load_db_user_help = "Load the instance with an updated db user."
+load_db_password_help = "Load the instance with an updated db password."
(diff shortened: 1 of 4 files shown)


falexwolf:
Thank you, Fred! I think the basic UX should work like this. One might think a bit through the consistency of passing a single `--db` arg vs. passing two `--db-user` and `--db-password` args, but I think this could work.

Code quality: This is a serious amount of code for which we'll need fine-grained unit tests.

We should also think through naming choices of endpoints & number of requests made so that this runs fast & we'll not want to rename an endpoint 2 months from now.

So, I'd say, let's give it another refactor. I'll review in detail tomorrow!

fredericenard:
Do you mean passing something like `user:password`?

fredericenard:
This [PR](https://github.com/laminlabs/lnhub-rest/pull/253) should solve the RLS issue by introducing a new table to store db_user_id for a specific collaborator.

PR #523, opened 2023-09-17
🐛 Unlock load on error

Koncopd:
Not sure how to test.

### DIFF
--- a/lamindb_setup/_load_instance.py
+++ b/lamindb_setup/_load_instance.py
@@ -14,8 +14,10 @@ from .dev._settings_storage import StorageSettings
 from .dev._settings_store import instance_settings_file
 from .dev._hub_client import call_with_fallback
 from .dev._hub_crud import sb_select_account_handle_name_by_lnid
+from .dev.cloud_sqlite_locker import InstanceLockedException, unlock_on_exception
 
 
+@unlock_on_exception
 def load(
     identifier: str,
     *,
@@ -128,7 +130,7 @@ def load(
                 f"'{user_info['handle']}' (id: '{locked_by}', name:"
                 f" '{user_info['name']}')."
             )
-        raise RuntimeError(lock_msg)
+        raise InstanceLockedException(lock_msg)
     if storage is not None and isettings.dialect == "sqlite":
         update_root_field_in_default_storage(isettings)
     load_from_isettings(isettings)
--- a/lamindb_setup/_load_instance.py
+++ b/lamindb_setup/_load_instance.py
@@ -20,6 +20,11 @@ from .dev.cloud_sqlite_locker import (
 )
 
 
+# this is for testing purposes only
+# set to True only to test failed load
+_TEST_FAILED_LOAD = False
+
+
 @unlock_cloud_sqlite_upon_exception(ignore_prev_locker=True)
 def load(
     identifier: str,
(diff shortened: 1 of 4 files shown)


lamindb_setup/_load_instance.py
```diff
@@ -14,8 +14,10 @@
 from .dev._settings_store import instance_settings_file
 from .dev._hub_client import call_with_fallback
 from .dev._hub_crud import sb_select_account_handle_name_by_lnid
+from .dev.cloud_sqlite_locker import InstanceLockedException, unlock_on_exception
 
 
+@unlock_on_exception
```
falexwolf, line 20: Could you rename to `unlock_cloud_sqlite_upon_exception`?

falexwolf:
This is a smart & interesting solution!

Initially, I thought we could encapsulate logic more, but I see that this should work just fine.

One way of testing it is by setting an internal variable to something invalid. Say, the instance settings cache file name or another global variable.

I've done this, for instance, here: https://github.com/laminlabs/lamindb-setup/blob/52a58238ecc17bbcff34c0bfbeb223811c994973/tests/prod-only/test_switch_and_fallback_env.py#L31-L34

This code triggers loading connection parameters from S3 if the global variable raises an exception.

lamindb_setup/dev/cloud_sqlite_locker.py
```diff
@@ -190,3 +197,26 @@ def get_locker(isettings: InstanceSettings) -> Locker:
...
+        try:
+            return func(*args, **kwargs)
+        except Exception as exc:
+            if isinstance(exc, InstanceLockedException):
+                raise exc
+            # this is ugly but i don't see any other way to do it properly
+            # this is ignored because if not then the current instance could be unlocked
+            # during the try of loading a new instance
+            elif isinstance(exc, RuntimeError):
+                if str(exc).startswith("Currently don't support init or load"):
```
falexwolf, line 216: This is dangerous as it might easily change in the future.

falexwolf:
The decorator is good!

For reference, I've used a "callback function" for handling hub-related errors here: https://github.com/laminlabs/lamindb-setup/blob/52a58238ecc17bbcff34c0bfbeb223811c994973/lamindb_setup/dev/_hub_core.py#L158-L163

It's a completely analogous situation. The decorator might have been prettier also in my case, but I thought the "callback function" is a bit nicer to read because I'm also using return values.

lamindb_setup/_load_instance.py
Koncopd: Done.

lamindb_setup/dev/cloud_sqlite_locker.py
Koncopd: Changed logic there, removed this particular case and added the parameter that allows to ignore exceptions if locker is the same as the previous locker after the wrapped function execution.

lamindb_setup/_load_instance.py
```diff
@@ -128,7 +138,12 @@ def load(
                 f"'{user_info['handle']}' (id: '{locked_by}', name:"
                 f" '{user_info['name']}')."
             )
-        raise RuntimeError(lock_msg)
+        raise InstanceLockedException(lock_msg)
+
+    # this is for testing purposes only
+    if _TEST_FAILED_LOAD:
```
Koncopd, line 144: @falexwolf this is probably a bad idea, but i don't see how else to trigger an error exactly at the right place.

falexwolf: It's OK for now!

falexwolf:
It looks good to me!

Just to make sure: Did you also watch out for not deteriorating performance with additional requests to AWS S3 etc.? I don't think this is an issue, here.

Good to merge!

Koncopd:
Additional requests have to happen on exception, then `unlock` is called and only there. There is no way to unlock without these requests.

Issue #529, opened 2023-09-20
Centralize SQLite instance locking


falexwolf:
@Koncopd, why is this not centralized/modularized within `update_local_sqlite_file()`?

https://github.com/laminlabs/lamindb-setup/blob/5c06ab60f70d3cc4036d39d6379955faf3eec583/lamindb_setup/_load_instance.py#L128-L146

Koncopd:
Also the lock should probably be before the synchronization, not after. I will go through this and rework if it is the correct approach.

PR #530, opened 2023-09-20
♻️ Refactor lock

falexwolf:
- https://github.com/laminlabs/lamindb-setup/issues/529

### DIFF
--- a/lamindb_setup/dev/_settings_instance.py
+++ b/lamindb_setup/dev/_settings_instance.py
@@ -3,9 +3,13 @@ import shutil
 from pathlib import Path
 from typing import Literal, Optional, Set, Tuple, Union
 from uuid import UUID
-
+from .dev._hub_client import call_with_fallback
+from .dev._hub_crud import sb_select_account_handle_name_by_lnid
 from lamin_utils import logger
-
+from .cloud_sqlite_locker import (
+    InstanceLockedException,
+    EXPIRATION_TIME,
+)
 from ._settings_save import save_instance_settings
 from ._settings_storage import StorageSettings
 from ._settings_store import current_instance_settings_file, instance_settings_file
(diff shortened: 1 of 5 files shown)


Koncopd:
I would leave it for now as it is, without reordering, it might even pass but i am not sure we can really rely on the tests here. I will need to check manually before proceeding.

Koncopd:
Also you can remove the first lock, checking with `_check_sqlite_lock` is enough, it calls `lock()` if it wasn't called before, just remove this logging msg also https://github.com/laminlabs/lamindb-setup/blob/5c06ab60f70d3cc4036d39d6379955faf3eec583/lamindb_setup/dev/cloud_sqlite_locker.py#L173

falexwolf:
I'll test it a bit! I've thought through the tests & also this re-ordering & I'm pretty sure what I do here is sound.

I wouldn't want to touch the locker code itself though, that's all a bit magic to me. 😅 

If we can avoid that you have to go back, think through this, and spend time with this and instead can spend time working on other valuable stuff, like, e.g., storing notebooks or testing the product with ML use cases, then we make up some time.

lamindb_setup/dev/_settings_instance.py
```diff
@@ -111,10 +115,32 @@ def _update_local_sqlite_file(self) -> None:
                 "updating local SQLite & locking cloud SQLite (sync back & unlock:"
                 " lamin close)"
             )
+            self._cloud_sqlite_locker.lock()
```
Koncopd, line 118: This is not needed, `_check_sqlite_lock()` will call `lock()` under the hood.

falexwolf: In `.has_lock`? This seems very implicit.

falexwolf: I see:
image

Koncopd: yep there

falexwolf: I also think it makes sense logically: One first tries to lock but as this might not work without throwing an error one calls an explicit check.

falexwolf: I can read this code and easily understand it even though it's logic from you. I always think that's a good test for good code.

PR #533, opened 2023-09-22
✨ Manage db connection with vault

fredericenard:


lamindb_setup/dev/_hub_client.py
```diff
@@ -24,6 +24,10 @@ def load_fallback_connector() -> Connector:
 
 PROD_URL = "https://hub.lamin.ai"
 PROD_KEY = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZSIsInJlZiI6ImxhZXNhdW1tZHlkbGxwcGdmY2h1Iiwicm9sZSI6ImFub24iLCJpYXQiOjE2NTY4NDA1NTEsImV4cCI6MTk3MjQxNjU1MX0.WUeCRiun0ExUxKIv5-CtjF6878H8u26t0JmCWx3_2-c"  # noqa
+# "https://laminhub-rest-cloud-run-prod-xv4y7p4gqa-uc.a.run.app"
+PROD_HUB_REST_SERVER_URL = (
+    "https://laminhub-rest-cloud-run-staging-test-xv4y7p4gqa-uc.a.run.app"
```
falexwolf, line 29: This won't be white-listed.

falexwolf: We need to have it, for instance, at `https://api.lamin.ai`

lamindb_setup/_init_vault.py
```diff
@@ -0,0 +1,41 @@
+from typing import Optional
+
+from pydantic import PostgresDsn
+
+from ._settings import settings
+from .dev._hub_utils import LaminDsnModel
+from .dev.cloud_sqlite_locker import unlock_cloud_sqlite_upon_exception
+
+
+@unlock_cloud_sqlite_upon_exception(ignore_prev_locker=True)
```
falexwolf, line 10: We don't need the decorator here, because it's only relevant for SQLite instances!

lamindb_setup/_load_instance.py
```diff
@@ -132,6 +139,32 @@ def load(
...
+        if config_db_exists:
+            return get_db_from_vault_base(
+                vault_client=vault_client,
+                scheme=instance_result["db_scheme"],
+                host=instance_result["db_host"],
+                port=instance_result["db_port"],
+                name=instance_result["db_database"],
+                role=f'{instance_result["id"]}-{settings.user.uuid}-db',
+            )
+    except Exception:
```
falexwolf, line 163: This exception seems like a bad idea. Wouldn't we rather want to throw an error? If we enabled vault and we can't get credentials, we'll error downstream of this call, won't we?

lamindb_setup/_load_instance.py
```diff
@@ -132,6 +139,32 @@ def load(
...
+        pass
```
falexwolf, line 164: Also this. How would we trigger it?

pyproject.toml
```diff
@@ -41,6 +42,10 @@ dev = [
 hub = [
     "faker",
 ]
+vault = [
+    "sqlalchemy",
```
falexwolf, line 46: Why do we need sqlalchemy? 🤔

tests/test_vault.py
```diff
@@ -0,0 +1,246 @@
...
+)
+from lamin_vault.client.postgres._role_and_policy_exist import role_and_policy_exist
+from lamin_vault.utils._lamin_dsn import LaminDsn
+from lamindb_setup._load_instance import load
+from lamindb_setup._init_instance import init
+from lamindb_setup._init_vault import init_vault
+from lamindb_setup._settings import settings
+from psycopg2 import sql
+from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
+from sqlalchemy import create_engine
```
falexwolf, line 20: Introducing sqlalchemy just for this seems a bit much after we worked so hard to eliminate it from the dependencies :)

Why not just use psycopg2?

falexwolf (approved):
Super cool, Fred! In particular the tests seem very extensive 🤩 

I just have very small asks.

lamindb_setup/_init_vault.py
fredericenard: Ok

tests/test_vault.py
fredericenard: Yes you're right

lamindb_setup/_load_instance.py
fredericenard: I don't know, I wanted to avoid introducing a new SPOF with this as it requires making a call to `laminhub-rest` server. 

Here, in case in failure, we will just fallback to the logic we have without vault. 

Do you think it's better to raise an error and to make load fails in that case?

falexwolf: Oh, I didn't parse this!

I'd still actually raise the error. We now want to test using lamindb-setup in vault mode and I fear people won't even notice if things go wrong if there is no clear error (or, they notice in a weird unexpected downstream way).

If you really think you want this magic under-the-hood, could you add a comment that this fallback to the original logic?

tests/test_vault.py
falexwolf: Thanks, Fred!

PR #536, opened 2023-09-29
✨ Enabling server deployment from CLI

fredericenard:


falexwolf:
This is cool, but none of this is tested. Hence, we should consider it as experimental functionality.

fredericenard:
I put some basic unit tests where the endpoint is implemented [here](https://github.com/laminlabs/laminhub-rest/blob/main/tests/test_deployment.py).

Ideally, we would have an integration test that would deploy a server, test everything is fine and delete it. But it takes around 5 minutes to create the server...

What do you think we should do in that case?

PR #538, opened 2023-10-01
✨ Cache management

Koncopd:
Allows settings and clearing the cache folder
This is in progress

I don't like that it is managed through instance settings, but don't see any other way for now.

Also maybe we need a better migration for env files?

### DIFF
--- a/lamindb_setup/__main__.py
+++ b/lamindb_setup/__main__.py
@@ -21,6 +21,7 @@ register_help = (
 )
 track_help = "Track a notebook (init notebook metadata)."
 save_help = "Save a notebook."
+cache_help = "Manage cache."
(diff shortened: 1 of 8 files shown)


Koncopd:
Ok, this is more complicated than i thought.

This sets the cache dir for each instance separately. So we lose the setting on `lamin close`. And alos have to set this for each instance separately.
@falexwolf should this setting be in `current_user.env` then? Or add another env file like `storage_settings.env`?

falexwolf:
I think it's fine to have a cross-instance cache dir be written to `current_user.env`. Prettier and more line with the schema is likely the `storage_settings.env`, I guess. 🤔

Koncopd:
But again we will lose it upon user change.

PR #577, opened 2023-11-09
♻️  Refactor UPath

falexwolf:


Koncopd:
I don't think subclassing like you want here will really work due to this https://github.com/fsspec/universal_pathlib/blob/befb71ea1e1a8c3685ddea3828c99cfe172257a3/upath/core.py#L162

I did this ugly extending of upath exactly due to this `__new__` overwrite in `upath`.

Koncopd:
Can explain the problem in details in a call.

falexwolf:
Hm, so far things seem to work. A `__new__` shouldn't prevent subclassing. I've subclassed classes with `__new__` many times before. 🤔 
image

falexwolf:
Well, in any case, in making `UPath` a user-facing first-class citizen I'd like to thoroughly clean up documentation and understand the underlying logic. I'll play around with things a bit more.

If you're still available for a call, happy to call, of course!

falexwolf:
OK, I see that things seem to get a bit complicated...

falexwolf:
Fun:
image

(Because __new__ returns an S3Path, which is no longer a subclass of the LaminUPath then. Hm)

falexwolf:
Going roughly back to what you had, @Koncopd, just that it seems I was able to fix the docs somehow through fixing a few docstrings.

The whole class didn't display correctly before.

Koncopd:
> Fun: image
> 
> (Because **new** returns an S3Path, which is no longer a subclass of the LaminUPath then. Hm)

Yes, exactly, this is the problem here with the logic of `UPath.__new__`.

PR #581, opened 2023-11-14
🔧 Detect both missing and ahead migrations

sunnyosun:
Screenshot 2023-11-14 at 15 13 40


lamindb_setup/dev/django.py
```diff
@@ -20,18 +20,43 @@
 
 Otherwise, downgrade your Python library to match the database!
 """
+AHEAD_MIGRATIONS_WARNING = """
```
falexwolf, line 23: This here should be a format string.

falexwolf:
You're currently getting an error here because the db is not yet initialized:
image

PR #587, opened 2023-11-24
🚚 Centralize progressbar and convert to percentage

falexwolf:


Koncopd:
Can't say i like this approach, we don't need the progress bar on every download or upload, but it is forced now.

Koncopd:
Like we now have progress bar on every sqlite sync

falexwolf:
> Can't say i like this approach, we don't need the progress bar on every download or upload, but it is forced now.

Oh, apologies, I thought it's a minor refactor towards more code ergnomics, better user feedback and more aesthetics (percentage, Valentine was confused by the absence of a unit)

> Like we now have progress bar on every sqlite sync
Yes, I'm aware. I tested and I felt it made the waiting time more pleasant because I saw what was going on

Also Jeremie used the download_to and upload_to and I'm sure he would have wanted clearer logging (I speced out some code for him and missed it myself)

If you'd like to hide it again, happy to hide in paces. But the SQLite synching I'd keep it - we can also ask the team for this

Koncopd:
Ok, i am home and can expand a bit on the problems i see with this PR.
2 minor technical issues:

- No way to disable the progress bar now.
- The `callback` kwarg is completely blocked for `upload_to` and `download_to`.

But there is also a more important objection.
These functions were designed to be a thin utility layer on top of `upath` with pass-through arguments and no side effects. And i really expect them to be so when i work with storage. Now they have some very specific logic with fsspec callback inside. And it is not about purity, we will certainly just forget to change the things [like this](https://github.com/laminlabs/lamindb-setup/blob/7452d486ca631bfeb2c2c479930054b36e00b117/lamindb_setup/dev/upath.py#L156) when we need to change something zarr related, for example (who would guess that you need to go the lowest level of storage code we have for this?). If you remeber we had a discussion of mixing levels, this is still valid here.

Koncopd:
@falexwolf we can discuss keeping the progress bar on sqlite sync if needed. But we should check, i expect a lot of progress bar appearing out of nowhere now.

Koncopd:
If you are fine with it, i will move the logic related to the progress bars back to lamindb, but keep the classes related to progress bars here. I can also add the progress bar explicitly to the sqlite sync.

falexwolf:
> Ok, i am home and can expand a bit on the problems i see with this PR. 2 minor technical issues, easily fixable (will fix):
> [...]

All your points taken, and I don't have a strong opinion on them. It's perfectly fine with me to keep refactoring and fixing these minor issues.

> If you remember we had a discussion about mixing levels, this is still valid here.

Given that we didn't add new methods to UPath, unlike in our discussion, I thought this would be OK.

> we can discuss keeping the progress bar on sqlite sync if needed.

Yes, I think this is very helpful. My own feeling and the feedback from Valentine and others were that there is not enough responsiveness around waiting times.

> i expect a lot of progress bars appearing out of nowhere now.

This is one of the three things I wanted to achieve by refactoring the code. Whenever an explicit download or upload occurs, I thought we'd like some responsiveness. I agree that for very small files this is nonsense, but the sqlite file already seems to take some time.

> If you are fine with it, i will move the logic related to the progress bars back to lamindb, but keep the classes related to progress bars here. I can also add the progress bar explicitly to the sqlite sync.

I'm in principle fine with it, just regret the amount of time I invested and that you're going to invest. I felt the code is now better in the sense that I was able to remove several manual ad hoc definitions related to response, and fill in gaps that were there. So, simply reverting it isn't great either, I guess.

One could also add a flag `print_progress` to mute the progress bar in `download_to` and `upload_to`, and default it to False` and then take care to switch it on in sufficiently many places.

Koncopd:
> One could also add a flag print_progress to mute the progress bar in download_to and upload_to, and default it to False` and then take care to switch it on in sufficiently many places.
Ok, lets do like this.

PR #598, opened 2023-12-20
🔊 Warn on sync of non-existing paths

Koncopd:
subj

### DIFF
--- a/lamindb_setup/dev/upath.py
+++ b/lamindb_setup/dev/upath.py
@@ -168,7 +168,15 @@ def upload_from(self, path, print_progress: bool = False, **kwargs):
 def synchronize(self, filepath: Path, **kwargs):
     """Sync to a local destination path."""
     if not self.exists():
-        return None
+        warn_str = f"The original path {self} does not exist anymore"
+        if filepath.exists():
+            warn_str += (
+                f". However, the local path {filepath} still exists, you might want to"
+                " reupload the object back."
+            )
+        else:
+            warn_str += ", it is not possible to synchronize."
+        logger.warning(warn_str)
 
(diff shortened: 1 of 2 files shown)


lamindb_setup/dev/upath.py
```diff
@@ -168,6 +168,15 @@ def upload_from(self, path, print_progress: bool = False, **kwargs):
 def synchronize(self, filepath: Path, **kwargs):
     """Sync to a local destination path."""
     if not self.exists():
+        warn_str = f"The original path {self} does not exist anymore"
+        if filepath.exists():
+            warn_str += (
+                f". However, the local path {filepath} still exists, you might want to"
```
falexwolf, line 174: I would add a new line here, so that it's easy to read. Forget about the `.`

lamindb_setup/dev/upath.py
```diff
@@ -168,6 +168,15 @@ def upload_from(self, path, print_progress: bool = False, **kwargs):
...
+                " reupload the object back."
+            )
+        else:
+            warn_str += ", it is not possible to synchronize."
```
falexwolf, line 178: I think here, we should throw an error, shouldn't we?

falexwolf: In almost any use case, people would run into obscure downstream errors otherwise.

Koncopd: Yes, will change that.

Koncopd: added

lamindb_setup/dev/upath.py
Koncopd: done

PR #607, opened 2024-01-05
💚 Upgrade to Python 3.10

falexwolf:


Koncopd:
Something like this maybe https://github.com/laminlabs/nbproject/blob/f534da22a8650d2d9104bdfdfdbd2dcc4b371f78/pyproject.toml#L24C11-L24C11 ?

Koncopd:
And this https://github.com/pandas-dev/pandas/blob/2e218d10984e9919f0296931d92ea851c6a6faf5/setup.cfg#L34-L39

falexwolf:
Hmmm, trying to add more verbose output.

falexwolf:
The references are great, Sergei, but I fear it's something else.

Koncopd:
Pretty sure this is it. Check django reqs, it reuires python>=3.10.

Koncopd:
Look https://github.com/django/django/blob/5c043286e2fde48a6230cbd2d91d5bdd3c49a418/setup.cfg#L36

falexwolf:
But where do we need <3.10 then?

Koncopd:
`ERROR: Package 'django' requires a different Python: 3.9.18 not in '>=3.10'`
This exactly says it.

falexwolf:
But which package requires <3.10?

Koncopd:
Something like
"django>4.2,<4.3; python_version < '3.10'"
"django>=5.0,<4.2; python_version >= '3.10'"

falexwolf:
Wait, maybe it's just the worker?

Koncopd:
No, it is django that is being installed on python=3.9 and can't be because the latest version requires python>=3.10.

falexwolf:
That's what I mean. But the worker should be running 3.10 according to `build.yml`?

Koncopd:
btw the lock error happened because the test notebooks was run in 2 parallel CI runs at the same time.

PR #610, opened 2024-01-06
🚸 Add anonymous access (now works without login)

falexwolf:


lamindb_setup/_load_instance.py
```diff
@@ -47,7 +47,6 @@ def update_db_using_local(
                 db_dsn_local = LaminDsnModel(db=os.getenv("LAMINDB_INSTANCE_DB"))
             # read from a cached settings file
             elif settings_file.exists():
-                logger.important("loading db URL from local cache")
```
falexwolf, line 50: @sunnyosun - this is where I removed the overly verbose logging!

PR #611, opened 2024-01-08
🚸 Fix connection timeout

falexwolf:


Koncopd:
One downside of this is that it degrades the performance for local postgres, sqlite etc.

lamindb_setup/dev/django.py
```diff
@@ -73,6 +75,12 @@ def get_migrations_to_sync():
     return status, latest_migrs
 
 
+def close_if_health_check_failed(self) -> None:
+    if self.close_at is not None and time.monotonic() >= self.close_at:
+        self.close()
+    self.close_at = time.monotonic() + CONN_MAX_AGE
```
Koncopd, line 81: this should be only reset if it is not `None` i think.

Koncopd:
```
if self.close_at is not None:
  self.close_at = time.monotonic() + CONN_MAX_AGE
```

falexwolf: Agreed! Fixed below!

PR #612, opened 2024-01-08
🚸 Allow loading another instance in the same Python session

falexwolf:


falexwolf:
I don't fully understand the logic of this test but I assume it aimed for triggering the `RuntimeError` that I just consciously removed:
image

I'll comment it out, but @Koncopd, it'd be good if you took a look into this at some point.

Koncopd:
Do you think this is enough? I wonder what happens with schemas in this case.

falexwolf:
The main problem right now is caching: https://laminlabs.slack.com/archives/C04FPE8V01W/p1704760398942229

falexwolf:
Superseded by:

- https://github.com/laminlabs/lamindb-setup/pull/1077

PR #621, opened 2024-01-25
✨ Add storage client

falexwolf:


lamindb_setup/dev/_hub_crud.py
```diff
@@ -1,6 +1,26 @@
...
+    try:
+        response = client.functions.invoke("access-aws")
+    except (FunctionsRelayError, FunctionsHttpError) as exception:
+        err = exception.to_dict()
+        logger.warning(err.get("message"))
+    if response is not None:
+        credentials = response["Credentials"]
+        os.environ["AWS_ACCESS_KEY_ID"] = credentials["AccessKeyId"]
+        os.environ["AWS_SECRET_ACCESS_KEY"] = credentials["SecretAccessKey"]
+        os.environ["AWS_SESSION_TOKEN"] = credentials["SessionToken"]
```
Koncopd, line 21: As we discussed, it is better to pass the credentials to fsspec here https://github.com/laminlabs/lamindb-setup/blob/8a55c63fc42fdb631c6e2ecf6414e474865aedae/lamindb_setup/dev/_settings_storage.py#L69

falexwolf: Yes and thanks for linking the code, Sergei! This is my very first draft to play around with the experience. We'll do as you say.

falexwolf: One thing we also have to do is to introduce a fallback mechanism that makes it possible to access public buckets via `anon` in case the credentials don't have the policy for it.

falexwolf: https://laminlabs.slack.com/archives/C04TMJPNCBS/p1706049778482779

Koncopd: Ah, ok, I see. I can work on passing the credentials to s3fs and on the fallback when we can get them successfully from the hub.

But also would be geat to avoid the fallback if possible. I will check the  slack thread later.

PR #622, opened 2024-01-26
🎨 Move default storages to folders

falexwolf:
https://laminlabs.slack.com/archives/C04TMJPNCBS/p1706266454001749

### DIFF
--- a/docs/prod-only/test-cache-management.ipynb
+++ b/docs/prod-only/test-cache-management.ipynb
@@ -39,7 +39,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "ln_setup.init(storage=\"s3://lndb-setup-ci\", name=\"test-cache\")"
+    "ln_setup.init(storage=\"s3://lamindb-ci/test-cache\", name=\"test-cache\")"
    ]
   },
   {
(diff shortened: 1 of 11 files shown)


falexwolf:
@Koncopd, is there anything about the locker that assumes it's at the root of a bucket? Shouldn't it just work as well if the root is a folder within a bucket?

Koncopd:
Hm, it should work, yes.

falexwolf:
I'll merge this for now, because it blocks a sequence of things that I have to do to get the storage service up and running.

If there is a bug with the locker, we can fix it in another PR.

Koncopd:
I would rather prefer to figure out first what is the problem.

Koncopd:
I am checking now what is the problem here

Koncopd:
Ok, it looks like i have found the problem.

Koncopd:
In this notebook https://github.com/laminlabs/lamindb-setup/blob/folders/docs/prod-only/test-load-lock.ipynb
`root = UPath("s3://lamindb-ci/test-load-lock/", cache_regions=True)`
`exclusion = root / ".lamindb/_exclusion/"`
produced root `"s3://lamindb-ci/test-load-lock**//**.lamindb/_exclusion/"`
So the trailing slash in UPath is ab big problem actually, it results in double slash on `/` operator.

Koncopd:
Actually, i think it is a `UPath` bug, i will report it.

falexwolf:
Maybe it's fixed and we just need to upgrade upath?

falexwolf:
To get the tests entirely to work, I need to re-deploy a schema change in the hub, and then post-populate something. Because these changes are needed for anything else, I'll now merge them in this broken state to main.

Koncopd:
Maybe it is fixed indeed, i will check.

PR #626, opened 2024-01-27
🎨 Refactor and define hosted storage user journey

falexwolf:


lamindb_setup/dev/upath.py
```diff
@@ -410,16 +410,15 @@ def create_path(pathlike: Union[str, Path, UPath]) -> UPath:
         path = UPath(str(pathlike))  # UPath applied on Path gives Path back
     else:
         raise ValueError("pathlike should be of type Union[str, Path, UPath]")
-
     # remove trailing slash
     if path._parts[-1] == "":
         path._parts = path._parts[:-1]
 
-    if isinstance(path, S3Path):
-        path = UPath(path, cache_regions=True)
```
Koncopd, line 419: It is important to keep cache_regions, otherwise s3fs doesn't work with newly created instances.

lamindb_setup/dev/_settings_storage.py
```diff
@@ -50,25 +51,39 @@ def get_storage_region(storage_root: Union[str, Path, UPath]) -> Optional[str]:
...
+        from ._hub_core import init_storage as init_storage_hub
+        from ._hub_core import access_aws
+
+        uuid = init_storage_hub(ssettings)
+        logger.important(f"registered storage: {ssettings.root_as_str}")
+        ssettings._uuid = uuid
+        if storage == "create-s3":
+            access_aws()
+            logger.important(
+                "exported AWS credentials as env variables, valid for 12 hours"
```
Koncopd, line 85: Not 36 hours?

lamindb_setup/dev/upath.py
falexwolf: Yes, I've just temporarily commented out.

lamindb_setup/dev/upath.py
```diff
@@ -402,6 +396,30 @@ def inner(dir_path: Union[Path, UPath], prefix: str = "", level: int = -1):
...
+
+    def wrap_with_args(func):
+        # https://stackoverflow.com/questions/1782843/python-decorator-handling-docstrings
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            try:
+                return func(*args, **kwargs)
+            except Exception as exc:
+                if isinstance(args[0], S3Path):
+                    path = UPath(args[0], cache_regions=True)
```
Koncopd, line 410: It should always set cache_regions for s3, not only on exception. I would just keep the `create_path` function for this.

Koncopd: This is important because `create_path` will also allow all other connections we create with it further (in `lamindb` etc) also have it and prevent problems arising from not having this setting.

falexwolf:
Wow, this drives me pretty crazy.

The env variable approach actually worked when initializing the instance. But now, somehow, I have to pass credentials explicitly:

image

Koncopd:
Something is wrong, this always works.

Koncopd:
btw when you set the environment variables via `os.environ`, it only sets variables for the current python process, maybe this is the problem here.

Koncopd:
It is hard to say what happens from the screenshot alone.

Koncopd:
Just writing after our discussion: lazily setting the credentials might not work because already existing upaths don't update their credentials.

Koncopd:
I think you have to explicitly reset the credentials in this case.

Koncopd:
There is this function https://github.com/aio-libs/aiobotocore/blob/d36c373028fbb5969bb117089e5322d447ceb5de/aiobotocore/session.py#L85
let me think how to use it.

lamindb_setup/dev/_settings_storage.py
```diff
@@ -51,6 +53,43 @@ def get_storage_region(storage_root: Union[str, Path, UPath]) -> Optional[str]:
     return storage_region
 
 
+def create_path_with_credentials(path: UPath) -> UPath:
+    if not isinstance(path, S3Path):
+        return path
+    # the below is problematic, because it will assume that all subsequent
```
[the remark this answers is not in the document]
Koncopd: this is really important, this prevents a ton of problems.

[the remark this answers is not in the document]
falexwolf: Done! See below!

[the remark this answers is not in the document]
Koncopd: Also probably needs to restored in the original create_path, which is used further in lamindb. Or it is deleted?

lamindb_setup/dev/_settings_storage.py
```diff
@@ -50,25 +53,75 @@ def get_storage_region(storage_root: Union[str, Path, UPath]) -> Optional[str]:
...
+    # requests will then be treated as anon, and we'll never try something else
+    # if isinstance(path, S3Path) and not upath.AWS_CREDENTIALS_PRESENT:
+    #     # if we already know we don't have AWS credentials
+    #     # let's make an anon request from the get-go
+    #     return UPath(path, anon=True)
+    try:
+        path.fs.call_s3("head_bucket", Bucket=path._url.netloc)
+        upath.AWS_CREDENTIALS_PRESENT = True  # type: ignore
+    except NoCredentialsError:
+        # we don't have any credentials at all
```
Koncopd, line 70: Are you sure that this is needed? If there are no credentials but they can be received from the hub, then why not use them? Now it will only return anon.

Koncopd:
What is actually cached is this .fs object
image
This looks pretty problematic.

Koncopd:
seems to be `s3fs`, not `upath`
image

lamindb_setup/dev/_settings_storage.py
Koncopd: Also if AWS_CREDENTIALS_PRESENT logic is here and not in the original create_path, then relevant parts will break further in lamindb in the case of no credentials, because it uses create_path and assumes that it treats this case.

falexwolf: Thanks for pointing this out, Sergei! I will take care of this!

lamindb_setup/dev/_settings_storage.py
falexwolf: We can play with this in the future.

lamindb_setup/dev/upath.py
```diff
@@ -402,24 +383,57 @@ def inner(dir_path: Union[Path, UPath], prefix: str = "", level: int = -1):
...
+    path = UPath(path, cache_regions=True)
+    # the below is problematic, because it will assume that all subsequent
+    # requests will then be treated as anon, and we'll never try something else
+    # if isinstance(path, S3Path) and not upath.AWS_CREDENTIALS_PRESENT:
+    #     # if we already know we don't have AWS credentials
+    #     # let's make an anon request from the get-go
+    #     return UPath(path, anon=True)
+    try:
+        path.fs.call_s3("head_bucket", Bucket=path._url.netloc)
+        AWS_CREDENTIALS_PRESENT = True  # type: ignore
```
Koncopd, line 415: Not sure this makes sense. The point of AWS_CREDENTIALS_PRESENT is to avoid path.fs.call_s3 on every call of create_path. So if it is set, the anon check should not be done. This is how the original create_path function does this. Now create_path calls path.fs.call_s3 every time.

Koncopd: I think the correct logic is.
Make a global AWS_MANAGED_CREDENTIALS=None.
if neither AWS_MANAGED_CREDENTIALS nor AWS_CREDENTIALS_PRESENT  set, check with path.fs.call_s3. 
If fails, check for AWS_MANAGED_CREDENTIALS. If success, set AWS_MANAGED_CREDENTIALS=True and do the reuired logic. If fails, check for AWS_CREDENTIALS_PRESENT  and do the required logic with anon.
if both variables set, just return the path.
This looks the correct way to me. But maybe i just don't understand what needs to be achieved here.

Koncopd: Anyways the current variant doesn't look correct to me.

Koncopd: The point here is that this check should not be done for every UPath conversion with create_path in lamindb.

falexwolf: Please read the comment! I'll point you to it.

falexwolf: It's the comment in the code that ends 3 line above your GitHub comment :D

I tried to answer in this comment what you're pointing out here:
image

falexwolf: Let me know if the reasoning makes sense!

falexwolf: I think the repeated requests are OK because the canonical usage will center around the default storage.

I'm sure there is optimization potential here, but I think the main challenge will be to enable to work across different sets of credentials.

Koncopd: Not sure. This makes the head_bucket check every time create_path is called, and it is called a lot in lamindb.

Koncopd: But the whole thing is pretty hard to understand already as the PR has too many things and some them can affect the credentials logic also.

falexwolf: Do you see the basic problem that we'll never hit the `access-aws` check in case there are no credentials? That means: people without AWS credentials in their environment would never get credentials for our hosted storage?

falexwolf: That's why we can't do the naive AWS_CREDENTIALS_PRESENT check we did before in the same way anymore.

Koncopd: Yes, i see it. The problem is current logic though is that it won't update the root https://github.com/laminlabs/lamindb-setup/blob/a03dfe2f2970752e4392c6302631eb2b400a6f63/lamindb_setup/dev/_settings_storage.py#L173 in this case.

So what happens as i understand it when there are no credentials.
the root is set with anon=True and is not updated.

also setting the environment variables seems useless to me if `s3fs.S3FileSystem.cachable = False` is not set due to s3fs caching behaviour described in my comments.

falexwolf: Agreed about setting the environment variables being useless. But I also think it doesn't hurt right now and it's nice to test things (it's a nice overwriting behavior upon calling `access_aws`).

PR #629, opened 2024-01-29
🚚 Rename bionty to bionty-base

sunnyosun:


falexwolf:
Handling this here, otherwise I can't get the tests to work!

- https://github.com/laminlabs/lamindb-setup/pull/626

PR #634, opened 2024-01-31
♻️ Rework s3 access check

Koncopd:
https://www.notion.so/laminlabs/Manage-AWS-credentials-client-side-1c6601180dc04776aefc252c7fa25837

https://laminlabs.slack.com/archives/C04FPE8V01W/p1706743094933629

Doesn't work with managed credentials for now.

### DIFF
--- a/lamindb_setup/dev/_s3access.py
+++ b/lamindb_setup/dev/_s3access.py
@@ -4,6 +4,7 @@ import os
 
 
 class s3Access:
+    # no checks happen for buckets in this dict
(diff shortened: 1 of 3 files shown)


Koncopd:
@falexwolf interesting test failure in build / storage
i realize now that we wanted lazy root check and it is not actually lazy due to
```
lamindb_setup/dev/_settings_save.py:53: in save_instance_settings
    f.write(f"lamindb_instance_storage_root={str(**settings.storage.root**)}\n")
```

PR #638, opened 2024-02-02
✅ Implement proper integration tests for hosted storage

falexwolf:


lamindb_setup/dev/upath.py
```diff
@@ -397,10 +399,11 @@ def convert_pathlike(pathlike: Union[str, Path, UPath]) -> UPath:
 def create_path(path: UPath) -> UPath:
     global AWS_CREDENTIALS_PRESENT
 
+    print(f"create path: {path}")
     path = convert_pathlike(path)
     if not isinstance(path, S3Path):
         return path
-    path = UPath(path, cache_regions=True)
+    path = UPath(path)
```
Koncopd, line 406: I guess this just here because we wanted to test without it. Should be restored i believe.

falexwolf: Yes, I just restored it.

PR #644, opened 2024-02-04
♻️ Check manual credentials in path.fs, actually reuse anon

Koncopd:
subj

### DIFF
--- a/lamindb_setup/dev/upath.py
+++ b/lamindb_setup/dev/upath.py
@@ -4,7 +4,7 @@
 import os
 from datetime import datetime
 from datetime import timezone
-import botocore
+import botocore.session
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


Koncopd:
@falexwolf is hosted init broken again?

falexwolf:
Something seems off, yes. I haven't debugged it yet.

falexwolf:
It started to appear all of a sudden, I need to check this.

Koncopd:
It seems to be on the side of hub. Also the same test on `build / prod-staging (staging)` passes.

falexwolf:
Yes, so, this is the reason for why hosted storage was broken: https://laminlabs.slack.com/archives/C04TMJPNCBS/p1707171896355059

PR #661, opened 2024-03-05
♻️ Move generic hashing and storage utilities from `lamindb` into `lamindb-setup`

bpenteado:
- https://github.com/laminlabs/lamindb/pull/1476
- https://www.notion.so/laminlabs/Monitor-hosted-storage-usage-enable-safe-deletes-4afbd336989c4f578fc8fab2194328eb

### DIFF
--- a/lamindb_setup/core/upath.py
+++ b/lamindb_setup/core/upath.py
@@ -506,7 +506,6 @@ def get_stat_file_cloud(stat: Dict) -> Tuple[int, str, str]:
 
 def get_stat_dir_s3(path: UPath) -> Tuple[int, str, str, int]:
     import boto3
-    from lamindb_setup.core.upath import AWS_CREDENTIALS_PRESENT
 
     if not AWS_CREDENTIALS_PRESENT:
         # passing the following param directly to Session() doesn't
--- a/lamindb_setup/core/upath.py
+++ b/lamindb_setup/core/upath.py
@@ -9,7 +9,7 @@ from pathlib import Path
 from typing import Literal, Dict
 import fsspec
 from itertools import islice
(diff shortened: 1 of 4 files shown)


falexwolf:
Could you cross-link this with the work package?

bpenteado:
> Could you cross-link this with the work package?

Done!

lamindb_setup/core/upath.py
```diff
@@ -484,3 +485,65 @@ def create_path(path: UPath, access_token: Optional[str] = None) -> UPath:
         secret=credentials["secret"],
         token=credentials["token"],
     )
+
+
+def get_stat_file_cloud(stat: Dict) -> Tuple[int, str, str]:
```
falexwolf, line 490: We need tests for this, @bpenteado!

In lamindb, this was covered, now, I fear, it's uncovered by CI.

My guess is that there were no fine-grained unit tests in lamindb for this, but coverage came through more high-level functionality.

Given this is in lamindb-setup now and the interfaces are used by laminhub and lamindb, we need a more fine-grained test that reflects needs in both downstream repos (they should be the same if we design the code well).

falexwolf: Here is an issue for this: https://github.com/laminlabs/lamindb-setup/issues/665

tests/test_hashing.py
```diff
@@ -0,0 +1,35 @@
+import base64
```
falexwolf, line 1: I fear these tests don't executed - see the noxfile. I'll take care of this.

falexwolf: Fixed here: https://github.com/laminlabs/lamindb-setup/pull/666

PR #668, opened 2024-03-07
🏗️ Enable`ln.connect()` in lamindb

falexwolf:


falexwolf:
@Koncopd - I made an issue about this, it has nothing to do with the cache but between the impossibility of switching between instances: https://github.com/laminlabs/lamindb-setup/issues/669

I'm not entirely clear why `ln_setup.close()` enabled it before, it shouldn't have been possible because we currently can't switch between lamindb instances.

lamindb_setup/_check_setup.py
```diff
@@ -1,33 +1,39 @@
...
 
-def check_instance_setup(from_lamindb: bool = False):
-    # silence_loggers is called twice within the if conditions for latency
-    # reasons
+def check_instance_setup(from_lamindb: bool = False) -> bool:
+    from .core.django import IS_SETUP, setup_django
+
+    if _LAMINDB_CONNECTED_TO is not None:
+        return True
+    if IS_SETUP:
```
falexwolf, line 22: @Koncopd - this line makes it more rigorous because this measures whether django is setup or not.

falexwolf: This is why now, we need the `_test` flag. I will try this.

(Just confirming what you said on the call before!)

falexwolf: Tried it and it doesn't work because it doesn't attempt to access the SQLite file when run with `_test`. Let's see whether I can change this.

Koncopd: No, this should not matter, lock is called in `_load_db`, which is called always on load / connect.

PR #675, opened 2024-03-08
🎨 Make `get_stat_dir_s3` compatible with canonical storage policy (`s3:ListBucket` conditional on `prefix/`)

bpenteado:
Performance benchmarking: https://lamin.ai/laminlabs/lamindata/transform/krGp3hT1f78N5zKv


lamindb_setup/core/upath.py
```diff
@@ -506,36 +506,12 @@ def get_stat_file_cloud(stat: Dict) -> Tuple[int, str, str]:
...
-        from botocore import UNSIGNED
-        from botocore.config import Config
-
-        config = Config(signature_version=UNSIGNED)
-        s3 = boto3.session.Session().resource("s3", config=config)
-    else:
-        s3 = boto3.session.Session().resource("s3")
-    bucket, key, _ = path.fs.split_path(path.as_posix())
-    # assuming this here is the fastest way of querying for many objects
-    objects = s3.Bucket(bucket).objects.filter(Prefix=key)
```
falexwolf, line 533: This should be fast because it only makes a single network request.

lamindb_setup/core/upath.py
```diff
@@ -506,36 +506,12 @@ def get_stat_file_cloud(stat: Dict) -> Tuple[int, str, str]:
...
-    size = sum([object.size for object in objects])
+    objects = path.fs.find(path.as_posix(), detail=True)
```
falexwolf, line 509: This might be slow depending on the fsspec implementation.

lamindb_setup/core/upath.py
```diff
@@ -506,36 +506,12 @@ def get_stat_file_cloud(stat: Dict) -> Tuple[int, str, str]:
...
+    size = sum([object["size"] for object in objects.values()])
```
falexwolf, line 510: Does .values() make another network request?

lamindb_setup/core/upath.py
```diff
@@ -506,36 +506,12 @@ def get_stat_file_cloud(stat: Dict) -> Tuple[int, str, str]:
...
     md5s = [
         # skip leading and trailing quotes
-        object.e_tag[1:-1]
-        for object in objects
+        object["ETag"][1:-1]
+        for object in objects.values()
```
falexwolf, line 514: Why do we call `.values()` again here?

falexwolf (changes requested):
- Please add a test!
- Please run on a large folder and report the performance difference.

lamindb_setup/core/upath.py
Koncopd: s3fs uses `list_object_v2` here as we wanted https://github.com/fsspec/s3fs/blob/0fbefba934073830075e02f4c2246c3521a15b10/s3fs/core.py#L755

lamindb_setup/core/upath.py
Koncopd: I agree, there should not be 2 list traversals, but just one.

falexwolf:
Here is a directory from an LTS cellxgene census release that you can use for the performance check:

```
ln.UPath("s3://cellxgene-data-public/cell-census/2023-12-15/soma").view_tree(level=4)
```

falexwolf:
For testing, you could e.g. use this one:
```
ln.UPath("s3://lamindb-dev-datasets/iris_studies/study0_raw_images").view_tree()
```

lamindb_setup/core/upath.py
bpenteado: Fixed

lamindb_setup/core/upath.py
bpenteado: Fixed

lamindb_setup/core/upath.py
```diff
@@ -345,6 +345,7 @@ def inner(dir_path: Path, prefix: str = "", level: int = -1):
         f" {n_files} files{suffix_message}): {folder_tree}"
     )
     logger.print(message)
+    return message
```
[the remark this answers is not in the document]
falexwolf: If you make this change, you should also update the docstring.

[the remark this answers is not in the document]
bpenteado: I figured - was going to ping you on this one.

I'll leave it as it was (logging) and will capture the stdout in tests.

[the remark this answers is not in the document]
falexwolf: I think that's good, if we make it return a string, we shouldn't call it "view_...()"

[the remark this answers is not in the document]
falexwolf: I think you forgot to revert the change.

falexwolf (approved):
Just the thing about `view_tree()` - as you said, I'd let it keeping.

tests/storage/test_storage_stats.py
```diff
@@ -0,0 +1,14 @@
...
+    import re
+
+    message = UPath(
+        "s3://lamindb-dev-datasets/iris_studies/study0_raw_images", anon=True
+    ).view_tree()
+    num_objects_view_tree = re.search(r"(\d+) files", message).group(1)
+    _, _, _, num_objects_stat = get_stat_dir_s3(
+        UPath("s3://lamindb-dev-datasets/iris_studies/study0_raw_images", anon=True)
+    )
+    assert num_objects_stat == int(num_objects_view_tree)
```
falexwolf, line 14: Hmmmm. 🤔

falexwolf: Instead of using regex, cold you save `n_files` as a private attribute in `UPath` and then test for it?

I also suggest you rename it to `n_objects`.

lamindb_setup/core/upath.py
```diff
@@ -344,7 +344,7 @@ def inner(dir_path: Path, prefix: str = "", level: int = -1):
         f"{path.name} ({n_directories} sub-{directory_info} &"
         f" {n_files} files{suffix_message}): {folder_tree}"
     )
```
falexwolf, line 346: Please save `n_files` as `._n_objects` and then use this to test for it comparing with `get_stat_s3...`

falexwolf: Please also rename `n_files` to `n_objects` for consistency with the `Artifact` registry and other parts of the code.

bpenteado: @falexwolf comments observed. 

The way I saved `n_objects` as a private attribute in `UPath` doesn't look like great code - can you please check and provide suggestions on how to dynamically create attributes in a robust manner, if relevant?

falexwolf: It isn't great code but better than using regex!

We can discuss better design over time, we're struggling a bit with building on UPath for non trivial reasons.

This PR is good to merge!

lamindb_setup/core/upath.py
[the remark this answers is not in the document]
Koncopd: Why is this needed? View absolutely should not modify or add UPath properties.

lamindb_setup/core/upath.py
Koncopd: @falexwolf in my opinion modifying attributes of UPath just for a test is not appropriate.

PR #679, opened 2024-03-13
👷 Use test utilities from `laminhub-rest` during local tests

bpenteado:


noxfile.py
```diff
@@ -52,10 +52,19 @@ def install(session: nox.Session, group: str) -> None:
         session.run(*"git clone https://github.com/laminlabs/lamin-cli".split())
         session.run(*"pip install lamin-cli".split())
     elif group == "hub-local":
-        session.run(*"pip install -e .[aws,dev,hub]".split())
-        session.run(*"pip install ./laminhub/rest-hub".split())
+        session.run(*"pip install lamindb".split())
```
falexwolf, line 55: Ouch, this will get us into trouble, pulling a pypi lamindb version; we shouldn't do this, I'm pretty sure.

noxfile.py
```diff
@@ -52,10 +52,19 @@ def install(session: nox.Session, group: str) -> None:
...
+        if os.getenv("GITHUB_ACTIONS") is None:
```
falexwolf, line 56: Also this looks way too complicated, there gotta be a simpler solution for this.

tests/hub-local/conftest.py
```diff
@@ -1,49 +1,19 @@
 import os
```
falexwolf, line 1: This file looks great! 😄

falexwolf (approved):
Great!

PR #680, opened 2024-03-14
✨ Create gating function to check for empty storage location

bpenteado:


lamindb_setup/core/upath.py
```diff
@@ -548,3 +548,15 @@ def get_stat_dir_gs(path: UPath) -> Tuple[int, str, str, int]:
...
     return sum(sizes), hash, hash_type, n_objects
+
+
+def check_s3_storage_location_empty(path: UPath) -> bool:
+    objects = path.fs.find(path.as_posix(), detail=True)
+    n_objects = len(objects.values())
+    if n_objects == 0:
+        return True
+    else:
+        raise ValueError(
```
falexwolf, line 559: The way this function is called and typed should rather make it return False here than raising an error.

lamindb_setup/core/upath.py
```diff
@@ -548,3 +548,15 @@ def get_stat_dir_gs(path: UPath) -> Tuple[int, str, str, int]:
     n_objects = len(md5s)
     hash, hash_type = hash_md5s_from_dir(md5s)
     return sum(sizes), hash, hash_type, n_objects
+
+
+def check_s3_storage_location_empty(path: UPath) -> bool:
+    objects = path.fs.find(path.as_posix(), detail=True)
+    n_objects = len(objects.values())
```
falexwolf, line 555: Is this the fastest way to get the number of objects? Can you add a comment re alternatives?

tests/hub-local/test_storage_deletion_gate.py
```diff
@@ -0,0 +1,35 @@
+from lamindb_setup.core.upath import check_s3_storage_location_empty
+from laminhub_rest.orm._account_instance import sb_insert_collaborator
+from laminhub_rest.test.utils.instance import TestInstance
+from laminhub_rest.test.utils.user import TestUser
+from laminhub_rest.test.utils.access_aws import get_upath_from_access_token
+from laminhub_rest.utils._supabase_client import SbClientAdmin
+import pytest
+
+
+def test_check_s3_storage_location_empty(
```
falexwolf, line 10: Pretty! 😄

lamindb_setup/core/upath.py
bpenteado: Since this is supposed to be a gating function, I'll keep it as is (raising error if storage is not empty) and return None if the opposite holds. Will type it accordingly.

lamindb_setup/core/upath.py
bpenteado: I can't guarantee with 100% certainty here - I just know it's faster than leveraging a `boto3` session via `s3.Bucket(bucket).objects.filter(Prefix=key)` (based on our benchmark: https://lamin.ai/laminlabs/lamindata/transform/krGp3hT1f78N5zKv).

@Koncopd any additional thoughts here?

bpenteado: I just pushed a simple update that uses the same method with `detail=False`. I'm not sure if it improves performance, but I would expect it to.

PR #682, opened 2024-03-14
🚸 Enforce empty storage during instance deletion

bpenteado:


lamindb_setup/core/_hub_core.py
```diff
@@ -83,13 +83,29 @@ def _init_storage(ssettings: StorageSettings, client: Client) -> UUID:
...
     )
+
     if instance_account is not None:
+        # we need to make sure the default 0-byte object in the storage location has
+        # not been deleted to avoid permission errors from leveraging s3fs on an
+        # empty subdirectory
+        path = create_path(
+            instance_account["storage"]["root"], settings.user.access_token
+        )
+        path.fs.touch(str(path / TOUCH_FILE_PATH))
```
falexwolf, line 104: Please move this into its own function within `_settings_storage.py` and call it something like:
```
def mark_storage_root(root: UPathStr):
     is_initialized_key = ".lamindb/_is_initialized"
     path.fs.touch(str(path / is_initialized_key))
```

And add the comment section there and reuse it in both places?

lamindb_setup/core/_settings_storage.py
```diff
@@ -92,6 +93,14 @@ def init_storage(storage: UPathStr, region: Optional[str] = None) -> "StorageSet
         ssettings._description = f"Created as default storage for instance {uid}"
         ssettings._uuid = init_storage_hub(ssettings)
         logger.important(f"registered storage: {ssettings.root_as_str}")
+    if root.startswith("s3://"):
+        from .upath import create_path
+        from ._settings import settings
+
+        # we need to touch a 0-byte object in the storage location to avoid
+        # permission errors from leveraging s3fs on an empty subdirectory
+        path = create_path(root, settings.user.access_token)
```
falexwolf, line 102: For the user, this shouldn't be necessary because we already `create_path()` when constructing the `StorageSettings` object. 🤔 

And for the hub, we manage access to storage through global AWS creds anyway, hence, it shouldn't be needed there, too. 🤔

Koncopd: I agree, here `ssettings.root` should be used instead of initialization with `create_path`. Also only if `ssettings.is_cloud` and `storage == "create-s3"`

Koncopd: Also `settings.user.access_token` should not be passed directly to `create_path`, it is retrieved automatically there.

Koncopd: Also i don't actually like that it is done on the client side, because it forces to initialize root at the very beginning. Can't this be done on the hub side via global aws credentials?

bpenteado: > For the user, this shouldn't be necessary because we already create_path() when constructing the StorageSettings object.

This makes sense to me. I should just pass `settings.storage.root` to `mark_storage_root`.

> And for the hub, we manage access to storage through global AWS creds anyway, hence, it shouldn't be needed there, too.

This I don't fully understand. What do you mean by hub?

bpenteado: > This I don't fully understand. What do you mean by hub?

For the hub test suite you mean, correct?

Koncopd: Not settings.storage.root, but `ssettings.root` from the code above.

Koncopd: Does anything happen in the hub on triggering `init_storage_hub`?

tests/hub-local/test_hosted_instance_deletion_gate.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_hosted_instance_deletion_gate(run_id, s3_bucket):
+    test_instance = create_hosted_test_instance(
+        f"test_instance_{run_id}", s3_bucket.name
+    )
+
+    # Make sure 0-byte file is touched upon storage initialization
+    path = create_path(
+        test_instance.storage_root, access_token=settings.user.access_token
```
Koncopd, line 19: Do not pass `settings.user.access_token` manually.

bpenteado: This will become irrelevant if we just leverage `ssetings.root`?

Koncopd: No, it should not be done in general.

Koncopd: You need to pass it explicitly only when you need a token different from the token of a current user.

lamindb_setup/core/_settings_storage.py
bpenteado: > Does anything happen in the hub on triggering init_storage_hub?

Just a simple upsert to the storage table with fields from `ssetings`.

lamindb_setup/core/_hub_core.py
```diff
@@ -98,13 +97,11 @@ def delete_instance(instance_identifier: str) -> None:
...
-        path = create_path(
-            instance_account["storage"]["root"], settings.user.access_token
-        )
-        path.fs.touch(str(path / TOUCH_FILE_PATH))
+        root = instance_account["storage"]["root"]
+        mark_storage_root(root)
 
         # gate storage and instance deletion on empty storage location
-        check_s3_storage_location_empty(path)
+        check_s3_storage_location_empty(root)
```
Koncopd, line 104: This reads global credentials. If you want to get hosted credentials here, you need to use `create_path`.

Koncopd: This is why this fails for the hosted instance in the test, it doesn't have hosted credentials.

lamindb_setup/core/upath.py
```diff
@@ -550,7 +550,9 @@ def get_stat_dir_gs(path: UPath) -> Tuple[int, str, str, int]:
     return sum(sizes), hash, hash_type, n_objects
 
 
-def check_s3_storage_location_empty(path: UPath) -> None:
+def check_s3_storage_location_empty(path: UPathStr) -> None:
+    if not isinstance(path, UPath):
+        path = convert_pathlike(path)
     objects = path.fs.find(path.as_posix())
     n_objects = len(objects)
     if n_objects > 1:
```
Koncopd, line 558: i think it might be better to use size instead of n_objects. touched file has 0 size i think.

lamindb_setup/core/_hub_core.py
```diff
@@ -83,13 +83,26 @@ def _init_storage(ssettings: StorageSettings, client: Client) -> UUID:
...
         select_instance_by_owner_name,
         owner=owner,
         name=name,
     )
+
     if instance_account is not None:
+        # we need to make sure the default 0-byte object in the storage location has
+        # not been deleted to avoid permission errors from leveraging s3fs on an
+        # empty subdirectory
+        root = instance_account["storage"]["root"]
```
Koncopd, line 100: you need `root = create_path(instance_account["storage"]["root"])`

tests/hub-local/test_hosted_instance_deletion_gate.py
```diff
@@ -0,0 +1,40 @@
...
```
Koncopd, line 19: `access_token=settings.user.access_token` is not needed here

lamindb_setup/core/_settings_storage.py
```diff
@@ -54,6 +55,12 @@ def get_storage_region(storage_root: UPathStr) -> Optional[str]:
     return storage_region
 
 
+def mark_storage_root(root: UPathStr):
+    if not isinstance(root, UPath):
```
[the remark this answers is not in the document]
Koncopd: `root = convert_pathlike(root)` without the check is enough.

lamindb_setup/core/upath.py
Koncopd: So we don't really need to take the changing number of objects into account. btw i i haven't checked but what happens for sqlite 33 instances? They always have at least an sqlite file in the bucket.

bpenteado: Good point. We're going to have a problem with sqlite instances if we condition instance deletion on size being 0.

I think for this case we're necessarily going to have to check for the presence of the specific instance sqlite file based on its naming convention, no?

falexwolf: Yes

falexwolf: We also need to account for the `_lamindb_initialized` file, don't we?

lamindb_setup/core/_hub_core.py
```diff
@@ -83,13 +83,26 @@ def _init_storage(ssettings: StorageSettings, client: Client) -> UUID:
 
 
 def delete_instance(instance_identifier: str) -> None:
```
falexwolf, line 85: Can we have a docstring or extensive comment here that says what it does?

lamindb_setup/core/_settings_storage.py
```diff
@@ -92,6 +98,10 @@ def init_storage(storage: UPathStr, region: Optional[str] = None) -> "StorageSet
         ssettings._description = f"Created as default storage for instance {uid}"
         ssettings._uuid = init_storage_hub(ssettings)
         logger.important(f"registered storage: {ssettings.root_as_str}")
+    if root.startswith("s3://"):
+        # we need to touch a 0-byte object in the storage location to avoid
```
falexwolf, line 102: This comment should be with the mark_storage_root function, then we don't need this comment twice in the code base.

lamindb_setup/core/upath.py
```diff
@@ -560,5 +561,6 @@ def check_s3_storage_location_empty(path: UPath) -> None:
         # regardless of the credentials registered in the path (see
         # lamindb_setup/core/_settings_storage/init_storage).
         raise ValueError(
```
falexwolf, line 563: Can you please raise a custom error?

```
class InstanceNotEmpty(Exception):
    pass
```

tests/hub-local/test_hosted_instance_deletion_gate.py
```diff
@@ -0,0 +1,40 @@
...
+    object_string_paths = path.fs.find(path.as_posix())
+    assert len(object_string_paths) == 1
+    assert (
+        object_string_paths[0]
+        == f"{test_instance.storage_root.replace('s3://', '')}/{IS_INITIALIZED_KEY}"
+    )
+
+    # Make sure gating function blocks instance and storage deletion if
+    # storage location is not empty
+    path.fs.touch(str(path / "new_file"))
```
falexwolf, line 30: Please use the UPath API: `path.touch()`, this line is ugly as hell otherwise

falexwolf: 😆

tests/hub-local/test_hosted_instance_deletion_gate.py
```diff
@@ -0,0 +1,40 @@
...
+    instance_slug = f"{settings.user.handle}/{test_instance.name}"
+    with pytest.raises(ValueError):
+        delete_instance(instance_slug)
+
+    # Make sure instance and storage deletion is possible with empty storage
+    path.fs.rm(str(path / "new_file"))
```
falexwolf, line 36: Please use the UPath API (it's the same as pathlib): `path.unlink()`

lamindb_setup/core/upath.py
bpenteado: What is `_lamindb_initialized`? Is it always written to `.lamindb/` in the storage location?

lamindb_setup/core/_settings_storage.py
```diff
@@ -59,7 +59,7 @@ def mark_storage_root(root: UPathStr):
     # we need to touch a 0-byte object in the storage location to avoid
     # permission errors from leveraging s3fs on an empty subdirectory
     root = convert_pathlike(root)
-    root.fs.touch(str(root / IS_INITIALIZED_KEY))  # type: ignore
+    root.touch(str(root / IS_INITIALIZED_KEY))  # type: ignore
```
falexwolf, line 62: String casting should no longer be needed, now, meaning, you should be able to remove `str()`

tests/hub-local/test_hosted_instance_deletion_gate.py
```diff
@@ -27,13 +27,13 @@ def test_hosted_instance_deletion_gate(run_id, s3_bucket):
...
     # storage location is not empty
-    path.fs.touch(str(path / "new_file"))
+    path.touch(str(path / "new_file"))
     instance_slug = f"{settings.user.handle}/{test_instance.name}"
     with pytest.raises(ValueError):
         delete_instance(instance_slug)
 
     # Make sure instance and storage deletion is possible with empty storage
-    path.fs.rm(str(path / "new_file"))
+    path.unlink(str(path / "new_file"))
```
falexwolf, line 36: This should read:

```
(path / "new_file").unlink()
```

But ideally you make a new path for the new file!

Paths are pretty objects that should be used 😍

tests/hub-local/test_hosted_instance_deletion_gate.py
```diff
@@ -0,0 +1,40 @@
...
+    object_string_paths = path.fs.find(path.as_posix())
+    assert len(object_string_paths) == 1
+    assert (
+        object_string_paths[0]
+        == f"{test_instance.storage_root.replace('s3://', '')}/{IS_INITIALIZED_KEY}"
+    )
+
+    # Make sure gating function blocks instance and storage deletion if
+    # storage location is not empty
+    new_file = UPath(path / "new_file")
```
falexwolf, line 29: We wouldn't need to call the `UPath` constructor here, but it's also not an issue.

PR #695, opened 2024-03-28
✨ `upath.to_url()`

falexwolf:


lamindb_setup/core/upath.py
```diff
@@ -421,6 +425,54 @@ def view_tree(
     logger.print(message)
 
 
+# TODO: replace with something faster at some point
+# also consolidate with what's in _settings_storage
+def get_bucket_region(bucket_name) -> str:
```
Koncopd, line 430: Why is this needed? We already have a proper function for that https://github.com/laminlabs/lamindb-setup/blob/ef0f0fe29edb9dfabc7d51528ca0d7c90e32c5fb/lamindb_setup/core/_settings_storage.py#L29

Koncopd: Your new variant looks problematic to me because it doesn't account for the case when no credentials are present in the system.

falexwolf: Yes, I don't like this. It's AI-generated.

lamindb_setup/core/upath.py
```diff
@@ -421,6 +424,48 @@ def view_tree(
...
+    - upath: A UPath object representing an S3 path.
+
+    Returns:
+    - A string containing the public URL to the S3 object.
+    """
+    if upath.protocol != "s3":
+        raise ValueError("The provided UPath must be an S3 path.")
+    bucket = upath._url.netloc
+    key = "/".join(upath.parts[1:])
+    region = get_bucket_region(bucket)
```
Koncopd, line 462: s3fs fs should have this information, i would not use it like this, we should avoid creating new client randomly, we need to use the underlying s3fs fs clinet.

falexwolf: Fixed in the commit above!

Koncopd: There is also fs.url for that https://s3fs.readthedocs.io/en/latest/api.html?highlight=url#s3fs.core.S3FileSystem.url

falexwolf: Hmm, this is what I took from your code in StorageSettings. 🤔

falexwolf: - https://github.com/laminlabs/lamindb-setup/issues/696

PR #708, opened 2024-04-05
🐛 Yet another fix for parallel import

Koncopd:


falexwolf:
@Koncopd - don't you agree we need a test; otherwise, we'll keep missing things. I'm just finishing the whole typing issues, that's why I haven't started yet.

Koncopd:
Yep, absolutely agree that we need a test and that it should be in lamindb.

PR #723, opened 2024-04-20
⬆️ Migrate to upath 0.2.2

Koncopd:


Koncopd:
Tried to migrate and got a bug https://github.com/fsspec/universal_pathlib/issues/216

falexwolf:
Nice! This is much prettier.

PR #728, opened 2024-04-23
✨ Introduce `local_storage` mode for cloud instances

falexwolf:
https://www.notion.so/laminlabs/Data-sharing-through-hybrid-local-remote-LaminDB-instances-e01025ed0e3e4f06948c4390894f20a8

Replace the initial draft:

- https://github.com/laminlabs/lamindb-setup/pull/717

Needs:

- https://github.com/laminlabs/laminapp-ui/pull/605


Koncopd:
I guess local instances also should have local storage, but it should be the same as root. Right?

falexwolf:
> I guess local instances also should have local storage, but it should be the same as root. Right?

Yes! The naming choice is slightly confusing for local instances, but simpler than saying "hybrid" for cloud instances.

The point is to make it easy for users to keep their data local also when working with cloud instances.

They'll have a switch on the hub called "Keep data local" or "Default to local storage" or similar.

I'm pretty sure we'll not allow to have purely local instances on the hub going forward.

Koncopd:
It is not actually clear than what is the difference between local and cloud now. Because a local instance can change its root to cloud but the local storage stays the same i guess. So now every instance can have cloud storage and local storage. So there is not much difference i suppose.

falexwolf:
> It is not actually clear than what is the difference between local and cloud now. Because a local instance can change its root to cloud but the local storage stays the same i guess. So now every instance can have cloud storage and local storage. So there is not much difference i suppose.

You are correct. I'm simply trying to achieve a more convenient UX for users who want to default to local storage working on a common server. I'll give this yet another refactor because the UX is still confusing.

PR #739, opened 2024-04-28
✨ Proper progress bars for upload and download

Koncopd:


Koncopd:
Why is adding a simple progress bar for directories is so complicated?..

PR #753, opened 2024-05-06
💚 Fix tests

fredericenard:


tests/hub-cloud/test_connect_instance.py
```diff
@@ -44,11 +45,8 @@ def test_connect_after_revoked_access():
             ln_setup.settings.instance.storage.root_as_str
             == "s3://lamindb-setup-private-bucket"
         )
-        delete_collaborator(
-            "laminlabs",
-            "static-test-instance-private-sqlite",
-            ln_setup.settings.user._uuid,
-            admin_hub,
+        collaborator_handler.delete_by_name(
```
falexwolf, line 48: Can we turn this into `delete_by_slug`? 

So that the signature becomes the following?
```
"testuser2", "laminlabs/static-test-instance-private-sqlite"
```

fredericenard: Done!

falexwolf: Oh, thank you! That's a lot more intuitive now!

PR #759, opened 2024-05-10
🧐 Switch on sql debug mode in django

sunnyosun:


falexwolf:
Tests are failing because of HuggingFace non-responsiveness again.

PR #760, opened 2024-05-10
✨ Retrieve another instance schema

fredericenard:


fredericenard:
@sunnyosun can I remove it? I did this because you needed to retrieve instance chela

PR #782, opened 2024-06-24
♻️ Save schema to hub upon init

falexwolf:


lamindb_setup/_init_instance.py
```diff
@@ -269,8 +269,10 @@ def init(
                 "locked instance (to unlock and push changes to the cloud SQLite file,"
                 " call: lamin close)"
             )
-        # we can debate whether this is the right setting, but this is how
-        # things have been and we'd like to not easily break backward compat
+        if isettings.dialect != "sqlite":
+            from ._schema_metadata import update_schema_in_hub
+
+            update_schema_in_hub()
```
falexwolf, line 275: @fredericenard - the instance and everything in the hub are fully generated at this point.

Evidently, this function is called by the user who creates the instance, so the JWT should be correct.

Is there any other logic pertaining to RLS somewhere that might not yet be in-place?

PR #800, opened 2024-07-23
🐛 Fix Supabase auth thread that prevents exit

fredericenard:
Here are related Supabase issues:
- https://github.com/supabase-community/supabase-py/issues/494
- https://github.com/supabase-community/supabase-py/issues/451

I'm not entirely sure removing sign out makes sense everywhere, fell free to put it back if needed.

### DIFF
--- a/lamindb_setup/core/_hub_client.py
+++ b/lamindb_setup/core/_hub_client.py
@@ -59,11 +59,11 @@ class Environment:
 
 # runs ~0.5s
 def connect_hub(
-    fallback_env: bool = False, client_options: ClientOptions = ClientOptions()
+    fallback_env: bool = False, client_options: ClientOptions | None = None
 ) -> Client:
     env = Environment(fallback=fallback_env)
-    client_options.auto_refresh_token = False
-    client_options.persist_session = False
+    if client_options is None:
+        client_options = ClientOptions(auto_refresh_token=False, persist_session=False)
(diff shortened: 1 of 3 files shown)


Koncopd:
Interesting, is it the reason our CI frequently hangs?

Koncopd:
Reading the pinned threads, it looks to me that sign out should be left in the code and actually probably added at some other places.
Thank you very much for finding the root of the problem.

lamindb_setup/core/_hub_client.py
```diff
@@ -62,6 +62,8 @@ def connect_hub(
     fallback_env: bool = False, client_options: ClientOptions = ClientOptions()
 ) -> Client:
     env = Environment(fallback=fallback_env)
+    client_options.auto_refresh_token = False
+    client_options.persist_session = False
```
falexwolf, line 66: Ah, I see!

Koncopd: Without sign out supabase tried to reuse the session between environments and it looks like anon tokens for different environments have different format, so the session tried to reuse sessions between environments and it resulted in a n error.

Koncopd: There is a problem that a session never expires so it sill hang forever in the auth table if we don't sign out and don't configure expiration time for **sessions** (not for access tokens).

PR #814, opened 2024-08-05
🚚 Rename `RecordMeta` to `Registry`

falexwolf:
- https://github.com/laminlabs/lnschema-core/pull/414

### DIFF
--- a/lamindb_setup/_schema_metadata.py
+++ b/lamindb_setup/_schema_metadata.py
@@ -356,6 +356,8 @@ class _SchemaHandler:
         return self.to_dict(include_django_objects=False)
 
     def _get_modules_metadata(self):
+        from lnschema_core.models import Record, Registry
+
         all_models = {
             module_name: {
                 model._meta.model_name: _ModelHandler(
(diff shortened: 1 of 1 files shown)


lamindb_setup/_schema_metadata.py
```diff
@@ -356,21 +356,23 @@ def to_json(self):
...
-                if model.__class__.__name__ == "RecordMeta"
-                and model.__name__ not in ["Record", "ORM"]
+                if model.__class__.__name__ == "Registry"
+                and model.__name__ != "Record"
                 and not model._meta.abstract
                 and model.__get_schema_name__() == module_name
             }
             for module_name in self.included_modules
         }
+        assert all_models
```
falexwolf, line 374: @fredericenard @sunnyosun this is so that the problem from this morning doesn't happen again: it checks that there are at least some models in the schema

falexwolf: But looking at this, it's somewhat bad code anyway to do this with strings... I'll refactor.
image

fredericenard: @falexwolf I'm not sure this check is the right one, because the dictionary wasn't empty. 
What was empty is `all_models[module_name]`.

fredericenard: I would suggest something like:

```
assert all(bool(module_dict) for module_dict in all_models.items())
```

falexwolf: Sounds good, Fred! Can you amend the check?

PR #821, opened 2024-08-14
🐛 Do not delete existing storage records in hub on failed inits

Koncopd:
https://laminlabs.slack.com/archives/C07DB677JF6/p1723656079519069?thread_ts=1723618378.437019&cid=C07DB677JF6


falexwolf:
Interesting, I think this is correct. Man, it'd be nice to have a test for these situations, but well.

Koncopd:
I agree, but these things are kind of hard to test properly.

PR #822, opened 2024-08-15
🐛 Make `set_managed_storage()` safer

Koncopd:
https://laminlabs.slack.com/archives/C07DB677JF6/p1723623568991179?thread_ts=1723618378.437019&cid=C07DB677JF6

### DIFF
--- a/lamindb_setup/_set_managed_storage.py
+++ b/lamindb_setup/_set_managed_storage.py
@@ -31,8 +31,11 @@ def set_managed_storage(root: UPathStr, **fs_kwargs):
             "Can't add additional managed storage locations for instances that aren't managed through the hub."
         )
 
+    #    ssettings = init_storage(
+    #        root=root, instance_id=settings.instance._id, prevent_register_hub=True
+    #    )
     ssettings = init_storage(
-        root=root, instance_id=settings.instance._id, prevent_register_hub=True
+        root=root, instance_id=settings.instance._id, register_hub=True
     )
     if ssettings._instance_id is None:
         raise ValueError(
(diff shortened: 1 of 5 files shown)


lamindb_setup/_set_managed_storage.py
```diff
@@ -28,14 +29,22 @@ def set_managed_storage(root: UPathStr, **fs_kwargs):
...
         root=root, instance_id=settings.instance._id, register_hub=True
     )
     if ssettings._instance_id is None:
         raise ValueError(
             f"Cannot manage storage without write access: {ssettings.root}"
         )
+
+    try:
+        register_storage_in_instance(ssettings)
+    except Exception as e:
```
falexwolf, line 43: Can you catch the exact exception type? Or at least add a comment. It's very hard to understand this otherwise.

lamindb_setup/_set_managed_storage.py
```diff
@@ -28,14 +29,22 @@ def set_managed_storage(root: UPathStr, **fs_kwargs):
...
+        if hub_record_status == "hub_record_created" and ssettings._uuid is not None:
```
falexwolf, line 44: Also this deserves extensive commenting.

falexwolf:
This is good, but please take the time to write a test for it and add the comments. It's very fundamental.

We'll also still need to update RLS, right? It's not a security concern, only an integrity concern, hence less urgent.

lamindb_setup/_set_managed_storage.py
Koncopd: Hm, it should catch any error that happens during saving the storage in the instance, i would not try to narrow it.

lamindb_setup/_set_managed_storage.py
Koncopd: I will comment.

Koncopd:
Not sure what kind of test can be added here, i am not really sure it is a good idea to add something failing, touching hub every time, than deleting the record to the CI.

falexwolf:
It's totally OK to add some failing and touch the hub all the time. It can deal with millions of requests.

Koncopd:
Maybe even in hub-local?

Koncopd:
I will try to add some testing here.

Koncopd:
I am not worried about the number of requests, rather about adding some garbage to the hub, and it will happen every time something incorrect happens to this flow (in future code edits etc).

PR #833, opened 2024-09-02
♻️ Use auth.set_auth to set access_token

Koncopd:


Koncopd:
Very strange, for some reason set_auth is not exposed here https://github.com/supabase/auth-py/blob/main/supabase_auth/_sync/gotrue_client.py
but exposed here in the unused class with the same name https://github.com/supabase/auth-py/blob/main/supabase_auth/_sync/client.py
It looks like this is the correct method to set access_token, but for some reason it is not exposed in the correct class used in supabase client.

falexwolf:
Should no longer be needed, right?

Koncopd:
I will re-check.

PR #840, opened 2024-09-03
✨ Allow login with new API key

Koncopd:


Koncopd:
Ok, this adds
`login(api_key="some_api_key")`
Loads a user with API key and allows to renew an access token with the stored API key.

Koncopd:
@falexwolf are you fine with this?

falexwolf:
This looks great except that I don't fully get the `.env` file management. 🤔 

The main issue is that his will raise an error for everybody when it goes live. We have the same issue for the instance env file where we'll add the `api_url`.

I think it'd be worthwhile to *no longer* use pydantic to validate the env file but instead just open the text file and parse it so that doesn't destroy everybody's workflow.

Koncopd:
I will try to fix this behavior here.

falexwolf:
We're already *not* using pydantic to write settings.

So, you should be able to create a `read_settings` function similar to the below and remove pydantic from the dependencies here.

https://github.com/laminlabs/lamindb-setup/blob/bc17873eb9b8ece81f3bbea88448456b88d9b551/lamindb_setup/core/_settings_save.py#L34-L65

lamindb_setup/core/_settings_store.py
```diff
@@ -65,6 +65,7 @@ class UserSettingsStore(BaseSettings):
     email: str
     password: str
     access_token: str
+    api_key: Optional[str] = None
```
Koncopd, line 68: @falexwolf this is actually enough to avoid the validation error.
Do you think the rewrite of settings load is still needed? It is not hard to do, especially via `dotenv_values` like [here](https://github.com/laminlabs/lamindb-setup/blob/bc17873eb9b8ece81f3bbea88448456b88d9b551/lamindb_setup/core/_settings_storage.py#L204), i just think it is reasonable to offload loading and validation logic to `pydantic-settings`.

falexwolf: Ok! If users don't run into errors that's alright!

falexwolf:
Another thing: This should be aware of an environement variable `"LAMIN_API_KEY"` which then allows to bypass things.

Exactly like wandb do it: https://docs.lamin.ai/wandb

Koncopd:
I though that `"LAMIN_API_KEY"` should be for the CLI. Should `ln_setup.login()` try to read the env variable also?

falexwolf:
> I thought that "LAMIN_API_KEY" should be for the CLI. Should ln_setup.login() try to read the env variable also?

Yes. The CLI shouldn't add additional logic. If people call `ln.setup.login()` without args this should behave in the same way as `lamin login`.

lamindb_setup/_setup_user.py
```diff
@@ -40,50 +43,65 @@ def load_user(email: str | None = None, handle: str | None = None) -> str | None
...
-    key: str | None = None,
+    user: str | None = None, *, key: str | None = None, api_key: str | None = None
 ) -> None:
     """Log in user.
 
     Args:
         user: handle or email
-        key: API key
+        key: permanent API key
+        api_key: new API key
```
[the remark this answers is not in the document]
falexwolf: Agreed! You can call the new key "Beta API key" and the old key "API key".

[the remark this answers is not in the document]
falexwolf: And at some point we switch it to "Legacy API key" and "API key".

falexwolf:
Well I agree that the CLI should be responsible for the logic around the `input()` prompt.

Koncopd:
Yes, but CLI should still check if `"LAMIN_API_KEY"` is present to avoid calling `input`.

lamindb_setup/_setup_user.py
```diff
@@ -40,50 +44,68 @@ def load_user(email: str | None = None, handle: str | None = None) -> str | None
...
     Args:
         user: handle or email
         key: API key
+        api_key: Beta API key
     """
-    if "@" in user:
-        email, handle = user, None
+    if user is None and api_key is None:
+        if "LAMIN_API_KEY" in os.environ:
+            api_key = os.environ["LAMIN_API_KEY"]
```
Koncopd, line 62: @falexwolf added the logic for `"LAMIN_API_KEY"`

falexwolf: Great!

Koncopd: Ok, merging this then.

PR #844, opened 2024-09-05
✨ Add submodule for entry point functionality

ap--:
Hello Lamins,

This PR adds an `_entry_points` submodule with a `call_registered_entry_points` function, which can then be used in `lamindb` as well as `lamin-cli` to provide a hook that other packages can use to do additional setup that might be required for lamindb to function correctly within their custom infrastructure.

usage example xref: laminlabs/lamin-cli#68

Cheers,
Andreas 😃


falexwolf:
Thanks, Andreas! I'll fix the CI runs.

What test shall we add for this?

ap--:
Good question, since the test will at the end rely on lamindb, lamin-cli and lamindb-setup it would be great if you could advise where I should add tests ideally.

In every dependency? or only in lamindb?

falexwolf:
I get how it works for CLI calls. 

```
call_registered_entry_points("lamin-cli.setup", module_globals=globals())
```

I'll read up in the original issue for how it would be called for lamindb.

Would you want this to be automatically called upon importing lamindb?

falexwolf:
Regarding the general layout, I'm responding in the original issue.

falexwolf:
Ok, tests fail because the secrets aren't passed on because you made your PR from a fork and there doesn't seem to be a setting to enable it:

image

I'll make you a collaborator with write access, then you don't need to make forks anymore.

ap--:
perfect. I'll add tests and will reopen from a branch in laminlabs/lamindb-setup

PR #846, opened 2024-09-05
✨ Add `lamin_setup.on_import` entry point

ap--:
This PR replaces #844 and adds a `lamin_setup.on_import` entry point, that allows other packages to do some environment setup if needed for custom infrastructure.


falexwolf (approved):
This looks great!

PR #848, opened 2024-09-06
✨ Load instance settings with edge function

Koncopd:
https://github.com/laminlabs/laminhub/pull/1209

### DIFF
--- a/lamindb_setup/core/_hub_core.py
+++ b/lamindb_setup/core/_hub_core.py
@@ -367,7 +367,7 @@ def _connect_instance(
     return instance, storage  # type: ignore
 
 
-def _connect_instance_remote(
+def _connect_instance_new(
     owner: str,  # account_handle
     name: str,  # instance_name
     client: Client,
@@ -409,7 +409,7 @@ def _connect_instance_remote(
     return instance, storage  # type: ignore
 
 
-def connect_instance_remote(
+def connect_instance_new(
     *,
     owner: str,  # account_handle
     name: str,  # instance_name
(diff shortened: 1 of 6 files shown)


Koncopd:
@falexwolf i think this is ready, along this https://github.com/laminlabs/laminhub/pull/1209

lamindb_setup/core/_settings_store.py
```diff
@@ -49,6 +49,7 @@ def system_storage_settings_file():
 
 
 class InstanceSettingsStore(BaseSettings):
+    api_url: Optional[str] = None
```
falexwolf, line 52: We also need the `schema_id`, here, don't we?

For `laminr` ([WP](https://www.notion.so/laminlabs/Build-an-R-SDK-laminR-8efbfd00157b4efe91b0927730492c0d)).

falexwolf:
This looks great to me!

It'd be nice though if this was starting to go live for the lamin team so that we build confidence in it. We could check for whether a user handle is in `sunnyosun`, `Koncopd`, `zethson`, `falexwolf` and then call the new connect function instead of the old one. Can you make this change.

Note: For laminr, we'll need to document how to call this edge function from within R. But that's probably for Notion.

lamindb_setup/core/_hub_core.py
```diff
@@ -367,6 +367,61 @@ def _connect_instance(
     return instance, storage  # type: ignore
 
 
+def _connect_instance_remote(
```
falexwolf, line 370: Can you call it `_connect_instance_new()` instead?

And then we'll delete the existing `_connect_instance` function once we spend 1-2 weeks testing it.

falexwolf (changes requested):
Thanks, Sergei! 3 small changes, please.

lamindb_setup/core/_settings_store.py
Koncopd: Added.

Koncopd:
>It'd be nice though if this was starting to go live for the lamin team so that we build confidence in it. We could check for whether a user handle is in sunnyosun, Koncopd, zethson, falexwolf and then call the new connect function instead of the old one. Can you make this change.

This i will add to `ln.connect` in a new PR when the edge function is deployed.

falexwolf:
Ok! Agreed!

lamindb_setup/core/_hub_core.py
Koncopd: done

PR #851, opened 2024-09-11
✅ Use connect_instance_new for Koncopd, sunnyosun and falexwolf for local testing

Koncopd:
Needs this https://github.com/laminlabs/laminhub/pull/1248 deployed


Koncopd:
Recent PRs in this direction decrease coverage, but it will be compensated when we decide to replace old `connect_instance` with `connect_instance_new` (via edge).

PR #852, opened 2024-09-12
🔧 Use LAMINDB_DJANGO_DATABASE_URL instead of DATABASE_URL

Koncopd:


lamindb_setup/core/_settings_instance.py
```diff
@@ -333,6 +333,10 @@ def _check_sqlite_lock(self):
     @property
     def db(self) -> str:
         """Database connection string (URI)."""
+        if "LAMINDB_DATABASE_URL" in os.environ:
```
falexwolf, line 336: Shouldn't this be called `LAMINDB_INSTANCE_DB`, because that's what it's called in pydantic, which has the same behavior.

Koncopd: Uh, this is yet another env varibale here https://github.com/laminlabs/lamindb-setup/blob/db0d9a0b8121e188a0be01288bde7f399ab83d97/lamindb_setup/_connect_instance.py#L73 i wasn't aware

Koncopd: Then need to rework this a bit.

Koncopd:
Hm, i am not sure now that it is better to point to already existing `LAMINDB_INSTANCE_DB`. `LAMINDB_INSTANCE_DB` is actually a proper thing, that is validated. And this one (`LAMINDB_DATABASE_URL`) is just something that overwrites any connection whatever it is. They should not be conflated i believe. And also double overwrite will be confusing we use `LAMINDB_INSTANCE_DB` here.

@falexwolf maybe lets call it not `LAMINDB_DATABASE_URL` but something like `LAMINDB_DJANGO_DATABASE_URL` just to be sure that is not accidently used, because `DATABASE_URL` is really generic.

falexwolf:
I like your last solution!

Koncopd:
Ok, merging then.

PR #857, opened 2024-09-18
♻️ Overhaul `init` for more convenient use in laminhub

Koncopd:
https://github.com/laminlabs/laminhub/issues/1271

This moves all internal utility arguments to `kwargs` in `lamindb_setup.connect`.
Adds 2 additional arguments: `_user` and `_write_settings`.

1. `_user` allows to pass user settings (with `access_token`) to `connect` and they are used instead of `lamindb_setup.settings.user`.
2. `_write_settings` when set to `False` forces `connect` not to write any settings env files. It is set to `True` by default for now.

I believe this addresses point 1-2 in the linked issue.

### DIFF
--- a/lamindb_setup/_connect_instance.py
+++ b/lamindb_setup/_connect_instance.py
@@ -204,11 +204,8 @@ def connect(
 
     access_token: str | None = None
     _user: UserSettings | None = kwargs.get("_user", None)
-    if _user is None:
-        user = settings.user
-    else:
-        user = _user
-        access_token = user.access_token
+    if _user is not None:
+        access_token = _user.access_token
 
     try:
         owner, name = get_owner_name_from_identifier(slug)
@@ -238,6 +235,10 @@ def connect(
                 return "instance-not-found"
         if isinstance(isettings, str):
             return isettings
+        # after we checked that isettings is not a string
+        # to lock passed user in isettings._load_db()
+        # has no effect if None or if not cloud sqlite
+        isettings._locker_user = _user
         if storage is not None:
             update_isettings_with_storage(isettings, storage)
         isettings._persist(write=_write_settings)
@@ -280,7 +281,7 @@ def connect(
         #     # raised by django when the access is denied
         #     except ProgrammingError:
         #         pass
-        load_from_isettings(isettings, user=user, write_settings=_write_settings)
+        load_from_isettings(isettings, user=_user, write_settings=_write_settings)
     except Exception as e:
         if isettings is not None:
             if _write_settings:
--- a/lamindb_setup/_connect_instance.py
+++ b/lamindb_setup/_connect_instance.py
@@ -235,9 +235,9 @@ def connect(
                 return "instance-not-found"
         if isinstance(isettings, str):
             return isettings
-        # after we checked that isettings is not a string
-        # to lock passed user in isettings._load_db()
-        # has no effect if None or if not cloud sqlite
+        # at this point we have checked already that isettings is not a string
+        # _user is passed to lock cloud sqite for this user in isettings._load_db()
+        # has no effect if _user is None or if not cloud sqlite instance
         isettings._locker_user = _user
         if storage is not None:
             update_isettings_with_storage(isettings, storage)
--- a/lamindb_setup/_connect_instance.py
+++ b/lamindb_setup/_connect_instance.py
@@ -27,6 +27,7 @@ from .core.cloud_sqlite_locker import unlock_cloud_sqlite_upon_exception
 if TYPE_CHECKING:
     from pathlib import Path
 
+    from .core._settings_user import UserSettings
     from .core.types import UPathStr
 
 # this is for testing purposes only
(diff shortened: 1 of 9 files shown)


lamindb_setup/_connect_instance.py
```diff
@@ -196,6 +191,11 @@ def connect(
         storage: Load the instance with an updated default storage.
     """
     isettings: InstanceSettings = None  # type: ignore
+
+    kwargs.get("access_token", None)
+    _raise_not_found_error = kwargs.get("_raise_not_found_error", True)
+    _test = kwargs.get("_test", False)
```
Koncopd, line 197: @falexwolf are you fine with moving all internal arguments to `kwargs` in `connect` and `init`?

Koncopd:
@falexwolf @fredericenard this is ready for a review.

lamindb_setup/_connect_instance.py
falexwolf: Ok!

lamindb_setup/_connect_instance.py
```diff
@@ -208,21 +219,29 @@ def connect(
...
-            isettings = _connect_instance(owner, name, db=db)
+            isettings = _connect_instance(owner, name, db=db, access_token=access_token)
         except InstanceNotFoundError as e:
             if _raise_not_found_error:
                 raise e
             else:
                 return "instance-not-found"
         if isinstance(isettings, str):
             return isettings
+        # after we checked that isettings is not a string
```
falexwolf, line 238: The comment here is incomprehensible to me just because the wording is poor.

lamindb_setup/core/_hub_core.py
```diff
@@ -137,10 +138,14 @@ def init_storage(
 
 
 def _init_storage(
-    ssettings: StorageSettings, auto_populate_instance: bool, client: Client
+    client: Client,
```
falexwolf, line 141: `client` has to be the last argument I believe. Otherwise the `call_with_fallback_auth` logic doesn't work, does it?

lamindb_setup/core/_hub_core.py
```diff
@@ -137,10 +138,14 @@ def init_storage(
...
+    ssettings: StorageSettings,
+    auto_populate_instance: bool,
+    created_by: str | None = None,
```
falexwolf, line 144: Can you call this `created_by_uuid` and type it with `UUID` instead of `str`? The conversion via `.hex` should come right before hitting the REST API so the the code internals are all properly typed.

lamindb_setup/core/_hub_core.py
```diff
@@ -253,29 +258,39 @@ def _delete_instance(
...
+        _init_instance,
+        isettings=isettings,
+        account_id=account_id,
+        access_token=access_token,
+    )
 
 
-def _init_instance(isettings: InstanceSettings, client: Client) -> None:
+def _init_instance(
+    client: Client, isettings: InstanceSettings, account_id: str | None = None
```
falexwolf, line 281: Well, I guess it works if `client` is the first argument. Then let's keep it this way.

But `account_id` has to be a `UUID` not a `str`!

lamindb_setup/core/_settings_instance.py
```diff
@@ -424,16 +431,16 @@ def is_on_hub(self) -> bool:
...
-        assert self.name is not None
-
-        filepath = self._get_settings_file()
-        # persist under filepath for later reference
-        save_instance_settings(self, filepath)
-        # persist under current file for auto load
-        shutil.copy2(filepath, current_instance_settings_file())
-        # persist under settings class for same session reference
-        # need to import here to avoid circular import
+    def _persist(self, write: bool = True) -> None:
```
falexwolf, line 434: Can you add a docstring or make this more explicit by calling the arg `save_to_disk` or similar?

falexwolf (changes requested):
Great, Sergei! Only very small changes requested from my side.

lamindb_setup/core/_hub_core.py
Koncopd: It doesn't matter at all, `client` is passed with the argument name.

falexwolf: Ok!

Koncopd:
Thx, will address the comments.

lamindb_setup/_connect_instance.py
falexwolf: I thought about this again and what I'd prefer is to have a properly typed signature of a function called `connect_internal()` that's wrapped by the user-facing `connect()`.

This way we eliminate **kwargs. We really shouldn't have them anywhere unless those places where we can't come up with a different solution.

falexwolf: The user-facing `connect()` will be super minimal and simple after also incorporating this change:  https://github.com/laminlabs/lamindb-setup/issues/862

falexwolf: And then we'll also be ready to attack this: https://github.com/laminlabs/pfizer-lamin-usage/issues/25

lamindb_setup/core/_hub_core.py
Koncopd: Can i leave it `created_by`, but redo the typing? Because i think it is just redundant to do both, i.e. renaming to `created_by_uuid` and typing with `UUID`?

lamindb_setup/_connect_instance.py
Koncopd: I agree, i just think that this probably to do in another PR, it requires a lot of other things, its own tests etc that i don't want to do here.

falexwolf: Ok!

lamindb_setup/_connect_instance.py
Koncopd: edited

lamindb_setup/core/_hub_core.py
Koncopd: Now using `UUID`.

lamindb_setup/core/_hub_core.py
Koncopd: done

falexwolf (approved):

lamindb_setup/core/_settings_instance.py
Koncopd: done

Issue #868, opened 2024-09-27
Permission error when saving a new artifact on python-3.12


ap--:
This is a bit of an edge case since it's I think unsupported as of now by being an issue on 3.12, but I'm reporting here, so that you're aware:

When the following conditions are met:
- installing `lamindb` on python3.12 (installation succeeds)
- and saving an artifact to an instance with remote storage

a permission error is raised.

```
lamin save myfile.txt --descriptio "My File!"
[...]
PermissionError
```

The reason this seems to happen, is that in the code path below, local_filepath becomes a remote path due to lamindb_setup relying on undefined behavior.

https://github.com/laminlabs/lamindb-setup/blob/bef6ce521b0192825e00960e296f4d290780ba53/lamindb_setup/core/_settings_storage.py#L393-L404

What happens is that in line 399, the following happens:

```
❯ python3.12 -m venv venv
❯ source venv/bin/activate
❯ pip install universal-pathlib==0.2.2
Looking in indexes: https://pypi.org/simple
Collecting universal-pathlib==0.2.2
  Using cached universal_pathlib-0.2.2-py3-none-any.whl.metadata (25 kB)
Collecting fsspec>=2022.1.0 (from universal-pathlib==0.2.2)
  Using cached fsspec-2024.9.0-py3-none-any.whl.metadata (11 kB)
Using cached universal_pathlib-0.2.2-py3-none-any.whl (46 kB)
Using cached fsspec-2024.9.0-py3-none-any.whl (179 kB)
Installing collected packages: fsspec, universal-pathlib
Successfully installed fsspec-2024.9.0 universal-pathlib-0.2.2

❯ python
Python 3.12.6 (main, Sep  6 2024, 19:03:47) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import upath
>>> fp = upath.UPath("s3://mybucket/.lamindb/abcdef123456")
>>> cd = upath.UPath("/some/local/path/lamin")
>>> cd / fp
S3Path('s3:///some/local/path/lamin/s3://mybucket/.lamindb/abcdef123456')
```

In `universal-pathlib>=0.2.5` this was fixed, and joining a remote path to an localpath as in `localpath / remotepath` raises a `ValueError` because the protocols don't match, and remote paths are all absolute.

[...]

PR #874, opened 2024-09-30
🐛 Fix cache paths

Koncopd:
Fix https://github.com/laminlabs/lamindb-setup/issues/868

### DIFF
--- a/docs/hub-prod/test-cloud-sync.ipynb
+++ b/docs/hub-prod/test-cloud-sync.ipynb
@@ -312,8 +312,8 @@
    "source": [
     "time.sleep(1)\n",
     "cloud_file = dir_sync / \"file1\"\n",
-    "# update cloud timestamp, exist_ok=False needed due to truncate=not exist_ok in upath\n",
-    "cloud_file.touch(exist_ok=False)  \n",
+    "# update cloud timestamp\n",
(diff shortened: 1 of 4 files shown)


docs/hub-prod/test-cloud-sync.ipynb
ap--, line 1: shouldn't this just be?

```python
cloud_file.touch(exist_ok=True)  
```

i.e. do you really want the file to be potentially truncated to 0 bytes?

see: fsspec/universal_pathlib#262

Koncopd: This one yes.
Without truncate is not implemented error.

Koncopd: I can also do just `unlink` and `touch` ofc.

ap--: This is caught in universal-pathlib==0.2.5 though. Am I missing an edge case with the touch behaviour here?

https://github.com/fsspec/universal_pathlib/blob/5486fa3c30cbb3f79637252d980994a1d738b0f5/upath/core.py#L970-L980

Koncopd: Yes, yes, i mean in s3fs. But if i do via upath here it won't update the timestamp. This is the whole point of calling it here.

PR #879, opened 2024-10-02
♻️ Base `connect_hub()` off of edge function

Koncopd:


lamindb_setup/core/_settings_instance.py
```diff
@@ -56,6 +56,8 @@ def __init__(
         schema: str | None = None,  # comma-separated string of schema names
         git_repo: str | None = None,  # a git repo URL
         is_on_hub: bool | None = None,  # initialized from hub
+        api_url: str | None = None,
+        schema_id: UUID | None = None,
```
Koncopd, line 60: @falexwolf not sure. I have to add them here for saving the instance settings. But i am not sure if they should be here even as just private variables.

falexwolf: I think this is OK, Sergei!

falexwolf:
@Koncopd, can we merge this one?

Koncopd:
One more thing and i will merge.

PR #885, opened 2024-10-14
✨ Overwrite access_token after renewal

Koncopd:
Fix https://github.com/laminlabs/pfizer-lamin-usage/issues/34


Koncopd:
Ok, this is easy if we rewrite the whole user settings, it should be fine like this for now.

PR #889, opened 2024-10-18
✨ Set anon for specified public s3 buckets if aws credentials fail

Koncopd:
Re https://github.com/laminlabs/pfizer-lamin-usage/issues/54 https://github.com/laminlabs/pfizer-lamin-usage/issues/55
This checks if aws credentials are able to resolve `head_bucket` on a test bucket (lamindata) and if not then sets anon=True for a specified list of public buckets.
This checks with `head_bucket` only once on init.

### DIFF
--- a/lamindb_setup/core/_aws_credentials.py
+++ b/lamindb_setup/core/_aws_credentials.py
@@ -147,7 +147,7 @@ class AWSCredentialsManager:
                         root = "/".join(path.path.rstrip("/").split("/")[:2])
                     else:
                         # write the bucket for everything else
-                        root = path._url.netloc
+                        root = path.drive
                     root = "s3://" + root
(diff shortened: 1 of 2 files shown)


Koncopd:
Not really sure that this is a good solution (specifying some preferred public buckets), but don't see any other way to avoid costly checks.

ap--:
I think having an option to set custom storage_options as json for Storage in the hub should fix this and be a very general solution. You'd set `{"anon": true}` for S3 buckets where you know that you want anonymous access.

Other options might become useful too in the future.

And it's generic, so support for other fsspec filesystems is included too in case they are supported.

Koncopd:
Yes, i also thought about this, a bit more complicated but probably better.

Koncopd:
I am leaning to merging it until we have a better solution.

Issue #748, opened 2024-05-04
Hashes for individual files in google cloud


Koncopd:
`UPath.stat` for google cloud paths has both "etag" and "md5Hash". We need to add this to https://github.com/laminlabs/lamindb-setup/blob/6bda3d8bc6c47c7707a79554149e8dc6a534e40f/lamindb_setup/core/upath.py#L745
![image](https://github.com/laminlabs/lamindb-setup/assets/3065736/d1f6e54a-16f7-48cb-beda-5ae9ed40fce8)
There are some processing of these hashes, so i am not sure how to this correctly as i am not aware of why this processing even needed, didn't work with hashes.
Now we just ignore hashes for individual files in gcp, but not for folders, which is strange.

falexwolf:
We first need to understand the difference between md5Hash and ETag. On AWS, for files below 50MB, the ETag is the md5 hash in hex representation.

Here, it seems that there is a difference, and google use base64 representation.

falexwolf:
And yes: we need to store the hash for file-like artifacts on GCP.

falexwolf:
Oh, that's very interesting. I fear though that AWS doesn't support CRC32c. It looks better in every regard than md5...

Let's keep this issue open and see what we can do here in the future.

For the time being, we'd likely resort to md5.

PR #898, opened 2024-11-06
✨ Compute hash for files in Google Cloud Storage and Hugging Face

Koncopd:
Closes https://github.com/laminlabs/lamindb-setup/issues/748

### DIFF
--- a/lamindb_setup/core/hashing.py
+++ b/lamindb_setup/core/hashing.py
@@ -47,14 +47,14 @@ def hash_set(s: set[str]) -> str:
     return to_b64_str(hashlib.md5(bstr).digest())[:HASH_LENGTH]
 
 
-def hash_md5s_from_dir(hashes: Iterable[str]) -> tuple[str, str]:
+def hash_from_hashes_list(hashes: Iterable[str]) -> str:
     # need to sort below because we don't want the order of parsing the dir to
     # affect the hash
     digests = b"".join(
         hashlib.md5(hash.encode("utf-8")).digest() for hash in sorted(hashes)
     )
     digest = hashlib.md5(digests).digest()
-    return to_b64_str(digest)[:HASH_LENGTH], "md5-d"
+    return to_b64_str(digest)[:HASH_LENGTH]
 
 
 def hash_code(file_path: UPathStr):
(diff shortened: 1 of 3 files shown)


falexwolf:
Darn, I think we only have code find files via blob hash: https://github.com/laminlabs/lamindb/blob/5c90903debbb6b35b2e00fc6f1be2a6e82ebe629/lamindb/core/_sync_git.py#L56

It should be rather straightforward, however, to get the blob hash.

As you say the commit hash is no valid proxy and indeed dangerous.

Koncopd:
Also not given by `stat`.

image

I am disabling hashing for huggingface paths for now.

falexwolf:
I believe hashes are very important and I'd not disable it.

If there is no way to get the full hash you can compute a pseudo hash like retro does by hashing the size and the file path.

Koncopd:
Yes, but this is a different approach than we have now, i would not do it just for huggingface and leave for as we have now for everything else.
I agree that hashing is important, and here i enable it for files in google cloud. For huggingface it is certainly safer also to have no hashing rather than commit hashing.

falexwolf:
The pseudo hash will be good enough for many purposes because these hugging face datasets will rarely be renamed.

Of course there are also a wide range of cases in which only the size is a good hash candidate but then there are cases where one deals with many images etc. of the exact same size and one needs another ingredient.

If there is a micro-second resolved created_at timestamp that's invariant under renaming on huggingface that'd be the perfect candidate for combination with the size to form a good pseudo hash.

falexwolf:
> Yes, but this is a different approach than we have now, i would not do it just for huggingface and leave for as we have now for everything else.

We're already working with different hash types and code the type of the hash in `_hash_type`. Hence, I don't see an issue to use a pseudo hash just for hugging face.

The big danger is that idempotency is lost and people might duplicate ingestion of hugging face datasets which will lead to problems.

Koncopd:
As you see, there is also no timestamp on file level, only commit level.

falexwolf:
Oh, but there is a `blob_id` which is the blob hash! Sorry for not spotting this earlier. You can simply use it, I'm pretty sure of that.

falexwolf:
If the blob_id is 40char it's the SHA1 that we also compute in lamindb-setup here: https://github.com/laminlabs/lamindb-setup/blob/2e905fdb8977b2ca232b9324cad4205b9253c671/lamindb_setup/core/hashing.py#L60

You can also do a simple comparison for a small file and you'll likely find it's the same thing.

falexwolf:
You should re-encode the hex code to base64 and truncate to 22 char because that's the limit for the hash field.

Along the lines of this: https://github.com/laminlabs/lamindb-setup/blob/2e905fdb8977b2ca232b9324cad4205b9253c671/lamindb_setup/core/hashing.py#L92

Koncopd:
Yep, undersood.

PR #906, opened 2024-11-24
♻️ Refactor setup bionty sources

falexwolf:
Context:

- https://github.com/laminlabs/lamindb/pull/2207

With:

- https://github.com/laminlabs/bionty/pull/171

### DIFF
--- a/lamindb_setup/core/_setup_bionty_sources.py
+++ b/lamindb_setup/core/_setup_bionty_sources.py
@@ -1,8 +1,12 @@
 from __future__ import annotations
 
+import os
 from typing import TYPE_CHECKING
 
 from django.db.utils import OperationalError, ProgrammingError
+from lamin_utils import logger
+
+from ._settings import settings as setup_settings
 
 if TYPE_CHECKING:
     from ._settings_instance import InstanceSettings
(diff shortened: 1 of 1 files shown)


lamindb_setup/core/_setup_bionty_sources.py
```diff
@@ -63,10 +67,21 @@ def _get_currently_used(key: str):
     Source.objects.bulk_create(all_records, ignore_conflicts=True)
 
 
-def load_bionty_sources(isettings: InstanceSettings):
+def load_bionty_sources(isettings: InstanceSettings | None = None):
     """Write currently_used bionty sources to LAMINDB_VERSIONS_PATH in bionty."""
-    if "bionty" not in isettings.schema:
-        return None
+    if isettings is None:
+        if setup_settings._instance_settings is not None:
```
Koncopd, line 73: @falexwolf could you clarify this? `setup_settings._instance_settings` can be `None` even if it is a normal single instance mode but the load of the settings wasn't triggered. I am not sure when it is triggered exactly, but checking `setup_settings._instance_settings` instead of `setup_settings.instance` looks very unsafe to me.

Koncopd: I mean i understand of course that the load is triggered by `setup_settings.instance` call, i am just not sure that `setup_settings.instance` is always called somewhere before `load_bionty_sources`

falexwolf: > checking setup_settings._instance_settings instead of setup_settings.instance

The problem with calling `setup_settings.instance` is that it takes much runtime in case it errors (which is what happens in the hub). At the point of calling `load_bionty_sources()` is called the instance is setup already. We shouldn't write logic that calls `load_bionty_sources()` to implicitly trigger the setup of the instance.

There is a general problem here with the design that `setup_settings.instance` is attempting to load from the local settings file if it's not yet configured.

Today, I'd rather allow `setup_settings.instance` to be `None` before an instance is configured than erroring. I'm not sure whether when we'll get to replace the implicit logic with clear tests for `is None` but I thought this new way of writing the code anticipates a better solution.

So, I'd not call it "unsafe".

One issue still is that for compat with the hub some patterns are not as clear as they can be. We'll likely be able to simplify once we're able to remove `MULTI_INSTANCE` mode or refactor it more.

PR #909, opened 2024-12-04
⬆️ Unpin supabase

Koncopd:
Replace with a lower and upper bound.


falexwolf:
If you do this we'll have to test all versions; don't we? Why not pin a version that's on conda-forge? That'd be less risky.

PR #915, opened 2024-12-18
🐛 Prepare for deletion of the storage_id column from the instance table

Koncopd:
Always search default storage through the storage table.

### DIFF
--- a/lamindb_setup/core/_hub_core.py
+++ b/lamindb_setup/core/_hub_core.py
@@ -239,6 +239,9 @@ def _delete_instance(
             )
             root_string = storage_record["root"]
             # gate storage and instance deletion on empty storage location for
+            # normally auth.get_session() doesn't have access_token
+            # so this block is useless i think (Sergei)
+            # the token is received from user settings inside create_path
             if client.auth.get_session() is not None:
                 access_token = client.auth.get_session().access_token
             else:
@@ -251,7 +254,13 @@ def _delete_instance(
             check_storage_is_empty(
                 root_path, account_for_sqlite_file=account_for_sqlite_file
             )
-    _update_instance_record(instance_with_storage["id"], {"storage_id": None}, client)
+    # remove this code block after migrtion, i.e. deleting 'storage_id' column of 'instance'
+    try:
+        _update_instance_record(
+            instance_with_storage["id"], {"storage_id": None}, client
+        )
+    except APIError:
+        pass
     # first delete the storage records because we will turn instance_id on
     # storage into a FK soon
     for storage_record in storage_records:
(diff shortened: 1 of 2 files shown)


lamindb_setup/core/_hub_core.py
```diff
@@ -251,7 +254,13 @@ def _delete_instance(
             check_storage_is_empty(
                 root_path, account_for_sqlite_file=account_for_sqlite_file
             )
-    _update_instance_record(instance_with_storage["id"], {"storage_id": None}, client)
+    # remove this code block after migrtion, i.e. deleting 'storage_id' column of 'instance'
```
falexwolf, line 257: We can also remove it straight away if we just set the values to `None` in the hub. I'd actually prefer that.

Koncopd: I would prefer to keep it until we actually migrate.

falexwolf: Why? I don't see the reason.

It makes the code complicated and unreadable. There is a really nice potential for a refactor once we remove this because the whole backward forward coupling of instance and storage is then finally disentangled.

falexwolf: Because we're no longer populating this column the tests are already passing even without deleting anything on the hub.

We still have to go through with it though so that deletions of legacy instances don't start failing.

falexwolf: Made an issue for this: https://github.com/laminlabs/laminhub/issues/1748

lamindb_setup/core/_hub_crud.py
```diff
@@ -89,15 +87,22 @@ def select_instance_by_id_with_storage(
     instance_id: str,
     client: Client,
 ):
-    response = (
+    # this won't find an instance without the default storage
+    data = (
         client.table("instance")
-        .select("*, storage!instance_storage_id_87963cc8_fk_storage_id(*)")
```
falexwolf, line 94: Oh no, was this somehow forgotten before?

falexwolf: Ok, this is only used during deletion; so it's not a big deal.

https://github.com/laminlabs/lamindb-setup/blob/db43bf308beb98c6834360c2950168fc1bb37c9d/lamindb_setup/core/_hub_core.py#L217

Maybe it was re-introduced at a later point or something similar.

No user will hit it.

Koncopd: Yes, it was forgotten, but it is pretty specific thing, onlz called on deletion by id, so not really user facing.

falexwolf: It's great you spotted it!

PR #917, opened 2024-12-23
🐛 Fix init instance tests and add init of a local instance in cwd

Koncopd:
Fix https://github.com/laminlabs/lamindb/issues/2180
Also fix incorrect tests in test_init_instance.py


Koncopd:
Test coverage was higher due to an incorrect test setup, can't really restore it fully with the correct setup.

PR #921, opened 2024-12-31
🏗️ Integrate `lnschema-core` into `lamindb`

falexwolf:
See:

- https://github.com/laminlabs/lamindb/pull/2309

### DIFF
--- a/lamindb_setup/_init_instance.py
+++ b/lamindb_setup/_init_instance.py
@@ -43,7 +43,7 @@ def get_schema_module_name(schema_name, raise_import_error: bool = True) -> str 
 
 
 def register_storage_in_instance(ssettings: StorageSettings):
-    from lamidb.base.users import current_user_id
+    from lamindb.base.users import current_user_id
     from lamindb.models import Storage
 
     from .core.hashing import hash_and_encode_as_b62
--- a/lamindb_setup/_init_instance.py
+++ b/lamindb_setup/_init_instance.py
@@ -99,35 +99,6 @@ def register_user_and_storage_in_instance(isettings: InstanceSettings, usettings
         logger.warning(f"instance seems not set up ({error})")
 
 
-def reload_schema_modules(isettings: InstanceSettings, include_core: bool = True):
-    schema_names = ["lamindb"] if include_core else []
-    # schema_names += list(isettings.schema)
-    schema_module_names = [get_schema_module_name(n) for n in schema_names]
-
-    for schema_module_name in schema_module_names:
-        if schema_module_name in sys.modules:
-            schema_module = importlib.import_module(schema_module_name)
-            importlib.reload(schema_module)
-
-
-def reload_lamindb_itself(isettings) -> bool:
-    reloaded = False
-    if "lamindb" in sys.modules:
-        import lamindb
-
-        importlib.reload(lamindb)
-        reloaded = True
-    return reloaded
-
-
-def reload_lamindb(isettings: InstanceSettings):
-    log_message = settings.auto_connect
-    if not reload_lamindb_itself(isettings):
-        log_message = True
-    if log_message:
-        logger.important(f"connected lamindb: {isettings.slug}")
-
-
 ERROR_SQLITE_CACHE = """
 Your cached local SQLite file exists, while your cloud SQLite file ({}) doesn't.
 Either delete your cache ({}) or add it back to the cloud (if delete was accidental).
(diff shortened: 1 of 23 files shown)


lamindb_setup/_init_instance.py
```diff
@@ -42,8 +42,8 @@ def get_schema_module_name(schema_name, raise_import_error: bool = True) -> str
 
 
 def register_storage_in_instance(ssettings: StorageSettings):
-    from lnschema_core.models import Storage
-    from lnschema_core.users import current_user_id
+    from lamindb.models import Storage
```
Koncopd, line 45: I am afraid this is much slower, no?

falexwolf: Why would it be much slower? 

(1) Do you mean because of a few additional dependencies that lamindb has over lamindb_setup? If that's the reason, I'm pretty sure it should be negligible. If it's not negligible we ought to think through optimizing import time on lamindb.

(2) If something else: 🤔 -- no, I can only speculate that you're having the auto-connect behavior in mind. Also here there is no additional slow-ness because you need to have django setup (connection established, if you will) also in the case of lnschema_core.

Koncopd: I mean 1), yes. lamindb is much much heavier in terms of imports, and I don’t really think it can be improved. We definitely need to test the import times. In my opinion significant degradation of performance is not justified by more convenient code organization (and I am not even sure about convenience here). lamin-cli is already pretty slow and we shouldn’t make it even slower I think. 

In summary I think we need to check the performance carefully before going forward with this.

falexwolf: Even though I don't believe it, let's assume that lamindb is so much heavier that it leads to noteworthy degradation of import times. It's hard to believe that it'd ever degrade more than 300ms.

Now consider that `from lamindb.models import Storage` and all similar statements is a dynamic import in `lamindb-setup`. It has to be because django needs to be setup before it runs through.

All instances in which the dynamic import is being run are cases in which run time is a few seconds. The worst case scenario of 300ms then isn't nice, but is also no big deal.

I've played with this for the past hours and I don't notice any slow down. If anything, I notice that several parts of the UX are faster (because I disentangled some recursive imports).

> more convenient code organization (and I am not even sure about convenience here)

This is not at all about convenience. It's about:

1. achieving a mono repo
2. https://github.com/laminlabs/pfizer-lamin-usage/issues/75
3. no longer depending on an external Django (https://laminlabs.slack.com/archives/C04FPE8V01W/p1735587801694429)

2. and 3. are enabled through 1. -- I imagine that 1. will enable more _fundamental_ code organization improvements like 2. and 3. beyond easing operational complications.

Koncopd: I am not that optimistic about performance. I know that it is fine on Linux (and probably macOS), but on windows it is already inconveniently slow and probably will be even slower with these changes. On windows it looks like additional imports make a difference.

Re part 2, I don’t entirely get why 1 and 2 are really needed, but we can discuss of course.

Koncopd: > because I disentangled some recursive imports

maybe this will make a positive difference on windows also.

Koncopd: There are also systems with pretty slow import of modules, like Helmholtz compute cluster, we need to have them also in mind.

falexwolf: > I am not that optimistic about performance. I know that it is fine on Linux (and probably macOS), but on windows it is already inconveniently slow and probably will be even slower with these changes. On windows it looks like additional imports make a difference.

Interesting, I wasn't aware of this at all. What is inconveniently slow? Can you make an issue on the `lamindb` repo and we discuss there?

> Re part 2, I don’t entirely get why 1 and 2 are really needed, but we can discuss of course.

1. is a big discussion, which we should rather have in person. 2. is what Andreas wants, and I agree with him in particular because it'll enable even deeper improvements like 3.

> maybe this will make a positive difference on windows also.

I hope so!

Koncopd: >Interesting, I wasn't aware of this at all. What is inconveniently slow? Can you make an issue on the lamindb repo and we discuss there?

Actually I am pretty sure we discussed this once or twice but never did anything. I will make an issue with measured loading times vs Linux. Also maybe on Helmholtz cluster (I believe they have pretty standard setup for compute clusters).

Koncopd: Btw import times on Helmholtz cluster are pretty abysmal at time.

PR #923, opened 2025-01-02
⚡️ Add the option not to reload lamindb on connect

Koncopd:


Koncopd:
I assume the test fails because we need a migration on laminhub.

PR #926, opened 2025-01-07
✨ Correct hashing of etag for http links

Koncopd:


sunnyosun:
Thank you, Sergei! How can I backfill the hashes of artifacts already registered?

PR #927, opened 2025-01-08
🚚 Upgrade to lamindb 1.0

falexwolf:


Koncopd:
django 5 is not compatible with python 3.9

Koncopd:
But what are the problems? This sounds very suspicious to me.

Koncopd:
I would prefer not to drop before EOL. This can definitely frustrate some users, we need to check first.

falexwolf:
> This can definitely frustrate some users, we need to check first.

I know that a few users might not be happy about it. But I fear that without it, there is no lamindb 1.0 (I mean it'd be super hacky and ugly not to us db_default). 

And without lamindb 1.0 we're practically dead-locked.

So, I'd accept that small nuisance given that we'll then be in a position of operating disruption free.

Do you see any other solution?

Koncopd:
Why is `db_default` needed?

falexwolf:
Here is an important approval:

- https://github.com/laminlabs/pfizer-lamin-usage/issues/107#issuecomment-2578344798

> Why is db_default needed?

To declare defaults on the database level in a meaningful way. Several new fields error otherwise on the hub. And it should be declared on the database level anyway. So if we don't do it now, we have to do it in a year or so with the next migration.

Koncopd:
`storage` CI group should be switched to 3.10 from 3.9 i think

Koncopd:
I think I understand why the tests fail here, I can help if needed.

lamindb_setup/core/_settings_instance.py
```diff
@@ -465,6 +465,13 @@ def _init_db(self):
         setup_django(self, init=True)
         _check_setup.IS_LOADING = False
 
+        from lamindb.models import Space
```
Koncopd, line 468: I suggest to move it into init, because persists is also called in connect. I think this is problematic in connect because it slows down and also can error if lamindb is not reloaded in connect (cli call).

Koncopd: Ok, sorry, it is not in persists, please ignore.

falexwolf: It is in `_init_db`. That's only called during init (`init=True`). Isn't it?

falexwolf:
> I think I understand why the tests fail here, I can help if needed.

Yeah, help is welcome! :D

falexwolf:
I think this is likely a non-idempotent test that still has the old schema in a dangling database sqlite database file?

Koncopd:
I don’t think it has anything to do with sqlite. Sometimes something fails and the test creates an instance record but can’t set the corresponding storage record to default. Next CI run loads the instance instead of creating but can’t delete it because it can’t find the default storage.

I will add a proper cleanup for this.

lamindb_setup/core/_settings_instance.py
Koncopd: Yes, correct, my mistake, I thought it was in persists not in init_db.

Koncopd:
And init should be reworked so this situation doesn’t occur anymore, but it has nothing to do with this PR.

Koncopd:
Ok, i have changed the test to avoid this nd cleaned up the stale records.

falexwolf:
Thank you very much, Sergei! ❤️

lamindb_setup/core/_settings_instance.py
falexwolf: The problem is that the other to records that are created in lamindb-setup (User & Storage) are touched on connect. This here is only touched on init.

Sorry, I didn't see this before!

PR #936, opened 2025-01-11
💚 Fix CI

Koncopd:


Koncopd:
I need to create a new instance instead of `testuser1/static-test-instance-private-sqlite` with the root `s3://lamindb-setup-private-bucket`. I will do this after the migration is complete.

PR #941, opened 2025-01-14
🚸 Error on auto-connect

falexwolf:
Now throws an error:
```
% lamin migrate deploy                
❌ You're in auto-connect mode, please run: lamin settings auto-connect false
```

This likely wasn't necessary before we removed lnschema-core.

### DIFF
--- a/lamindb_setup/_migrate.py
+++ b/lamindb_setup/_migrate.py
@@ -83,6 +83,11 @@ class migrate:
             update_instance,
         )
 
+        if settings.auto_connect:
+            raise SystemExit(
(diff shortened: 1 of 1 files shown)


falexwolf:
Tests are actually passing. It's just displaying a failure on GitHub. We've had this before. 🤷

Koncopd:
What is the problem with `auto-connect`? It is very inconvenient to always switch auto-connect to migrate.

Koncopd:
Ok, found the reported issue.

Koncopd:
Ok, in principle i understand the process now, indeed no obvious way to fix without switching off `auto-connect`.

PR #942, opened 2025-01-14
🩹 migrate without setting auto-connect to False

Koncopd:
For context https://github.com/laminlabs/lamindb-setup/pull/941

### DIFF
--- a/lamindb_setup/_check_setup.py
+++ b/lamindb_setup/_check_setup.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import functools
 import importlib as il
 import os
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


falexwolf (approved):
Had no time to try it out but I trust you did.

I like this and we could clean up the same logic we have for this in Lamin-CLI at some point (not urgent, but this here is more concise than what we have there).

falexwolf (commented):
Wait, I think you might be able to refactor this a bit; there is similar logic also on _check_setup I believe with a similar global variable that we shouldn't have twice

falexwolf (approved):
And I see you're using it. All good. Sorry on phone.

Koncopd:
But cli just calls these commands https://github.com/laminlabs/lamin-cli/blob/main/lamin_cli/_migration.py . It works for cli also of course.

> Wait, I think you might be able to refactor this a bit; there is similar logic also on _check_setup I believe with a similar global variable that we shouldn't have twice

I use here exactly this global variable from _check_setup .

Koncopd:
Ok, merging then.

Issue #933, opened 2025-01-10
Is suppressing the threading error on Windows still needed.


Koncopd:
Investigate deletion of https://github.com/laminlabs/lamindb-setup/blob/06f4993cf9069da02cf7feccc6100e7eb03a1317/lamindb_setup/__init__.py#L86

Issue #37, opened 2022-09-11
Strange error in login on Windows


Koncopd:
![image](https://user-images.githubusercontent.com/3065736/189534770-37034541-f47c-4034-8799-b274c33ddfb5.png)
Strange error on Windows that doesn't seem to break anything actually.
It happens here https://github.com/laminlabs/lndb-setup/blob/cb528b3732888c2408282ff13e67b4d930cb2338/lndb_setup/_hub.py#L79
because of `supabase` library, or more specifically, in imported `gotrue` package here https://github.com/supabase-community/gotrue-py/blob/46668453b0fbb0411af78f171b0dae80c5e2c23f/gotrue/_sync/client.py#L520
https://github.com/supabase-community/gotrue-py/blob/46668453b0fbb0411af78f171b0dae80c5e2c23f/gotrue/_sync/client.py#L616
https://github.com/supabase-community/gotrue-py/blob/46668453b0fbb0411af78f171b0dae80c5e2c23f/gotrue/_sync/client.py#L639
Setting larger `threading.TIMEOUT_MAX` didn't help.

However, this error doesn't break anything, it seems.

Probably should report to `supabase`.

PR #944, opened 2025-01-16
🔥 Remove thread error suppression on Windows

Koncopd:
We set `auto_refresh_token=False` for `supabase` anyways so it is redundant.

Closes https://github.com/laminlabs/lamindb-setup/issues/933

Context https://github.com/laminlabs/lamindb-setup/issues/37


falexwolf:
Nice clean up! 😄

PR #949, opened 2025-01-17
📌 Update aiobotocore upper bound to 2.17.0

fredericenard:


Koncopd:
Looks like it is not needed.

PR #961, opened 2025-02-04
✨ Extend schema json

fredericenard:
- We need a robust way to flag user defined link tables (see [this](https://laminlabs.slack.com/archives/C04884ZFESZ/p1738661169855089?thread_ts=1738575163.387139&cid=C04884ZFESZ) slack thread + [private discussion](https://laminlabs.slack.com/archives/D03N3RXHM5Y/p1738692463559449?thread_ts=1738691996.093749&cid=D03N3RXHM5Y) with @falexwolf)
- [This](https://laminlabs.slack.com/archives/C04884ZFESZ/p1728463675843879?thread_ts=1728396641.887319&cid=C04884ZFESZ) discussion with @Zethson made me realize some fields are missing in the schema json (primary key flag, maybe also the field used for default ordering and title displayed in the UI).


lamindb_setup/_schema_metadata.py
```diff
@@ -156,19 +157,23 @@ class FieldMetadata(BaseModel):
...
+        from lamindb.models import LinkORM
+
         self.model = model
         self.class_name = model.__name__
         self.module_name = module_name
         self.model_name = model._meta.model_name
         self.table_name = model._meta.db_table
         self.included_modules = included_modules
         self.fields = self._get_fields_metadata(self.model)
+        self.is_link_table = issubclass(model, LinkORM)
```
falexwolf, line 169: I thought we had this line already since the beginning. Good that we track this now!

falexwolf:
Great! 😄 

Can you please also add a test for a link table example?

And: can we also track whether a field is `editable` or not to display "edit" buttons on the UI? E.g. here: https://github.com/laminlabs/lamindb/blob/8b748245c200bfa00f3cf8d969d4f097c7615f20/lamindb/models.py#L842 and 
https://github.com/laminlabs/lamindb/blob/8b748245c200bfa00f3cf8d969d4f097c7615f20/lamindb/models.py#L853

lamindb_setup/_schema_metadata.py
fredericenard: We had this at the field level, to know if a field comes from a link table, but not at the model level.

tests/hub-local/test_update_schema_in_hub.py
```diff
@@ -63,6 +68,8 @@ def test_update_schema_in_hub(setup_instance):
         "model_name": "artifact",
         "schema_name": "core",
         "is_link_table": False,
+        "is_primary_key": False,
+        "is_editable": False,
```
fredericenard, line 72: @falexwolf It became not editable after I updated the PR, does it make sense?

falexwolf: Yes, many fields are now no longer editable.

fredericenard:
Should we merge?

PR #965, opened 2025-02-10
✨ Allow registering read-only storages

Koncopd:


lamindb_setup/core/_settings_storage.py
```diff
@@ -159,6 +160,10 @@ def init_storage(
                 created_by=created_by,
                 access_token=access_token,
             )
+    # we don not need to check that the storage is writable for read-only storages
```
Zethson, line 163:
```suggestion
    # we do not need to check that the storage is writable for read-only storages
```

Koncopd: @Zethson thank you.

PR #995, opened 2025-03-19
🐛 Fix rename of a legacy sqlite file and refactor space registration on init

Koncopd:


lamindb_setup/_init_instance.py
```diff
@@ -95,11 +95,16 @@ def register_user(usettings):
         pass
 
 
-def register_user_and_storage_in_instance(isettings: InstanceSettings, usettings):
-    """Register user & storage in DB."""
+def register_initial_objects(isettings: InstanceSettings, usettings):
```
falexwolf, line 98: Maybe call this register_initial_records()? We never use "objects" for records.

Koncopd: done

lamindb_setup/_init_instance.py
```diff
@@ -95,11 +95,16 @@ def register_user(usettings):
...
+    """Register space, user & storage in DB."""
     from django.db.utils import OperationalError
+    from lamindb.models import Space
 
     try:
+        Space.objects.get_or_create(
```
falexwolf, line 104: This is bad because it'll assign a random uid. The other logic we have assigns a uid that indicates that the "All" space is special.

Koncopd: it was always so, hm, nothing new here.

falexwolf: Hmmm, I see you're just moving it.

falexwolf: When was this introduced? I remember something that would populate `uid="000000000000"` (12 0s).

lamindb_setup/core/_settings_instance.py
```diff
@@ -498,7 +491,8 @@ def _load_db(self) -> tuple[bool, str]:
                     f"The SQLite file is being renamed from {legacy_file} to {self._sqlite_file}"
                 )
                 legacy_file.rename(self._sqlite_file)
-            return False, f"SQLite file {self._sqlite_file} does not exist"
+            else:
```
falexwolf, line 494: Oh, got it, makes sense!

lamindb_setup/_init_instance.py
Koncopd: i can add this.

Koncopd: but it won't have a random uid, it will have exactly `"00000000"`. See https://github.com/laminlabs/lamindb/blob/01abda450218a9647b1f23b5368193425a0fcf7a/lamindb/models/record.py#L896

Koncopd: It has been always like this, no changes. Also note that there is no auto-increment on space `uid`, so to create anew space `uid` has to be manually provided like i do here https://github.com/laminlabs/lamindb/blob/01abda450218a9647b1f23b5368193425a0fcf7a/tests/permissions/scripts/setup_access.py#L5

lamindb_setup/core/_settings_instance.py
falexwolf, line 494: Oh, got it, makes sense!

lamindb_setup/_init_instance.py
falexwolf: All good then!

falexwolf: Perfect.

falexwolf: > Also note that there is no auto-increment on space uid

It shouldn't, it should be a random uid created in central

falexwolf: > so to create anew space uid has to be manually provided like i do here https://github.com/laminlabs/lamindb/blob/01abda450218a9647b1f23b5368193425a0fcf7a/tests/permissions/scripts/setup_access.py#L5

This is not how it should be used but for testing it's OK. This should be a random uid.

falexwolf (approved):

lamindb_setup/_init_instance.py
Koncopd: Ok, got it.

PR #996, opened 2025-03-19
♻️ Remove leading underscore prefix in `lamin.db` filename

falexwolf:


lamindb_setup/core/_settings_instance.py
```diff
@@ -298,7 +298,7 @@ def schema(self) -> set[str]:
     @property
     def _sqlite_file(self) -> UPath:
         """SQLite file."""
-        filepath = self.storage.root / ".lamindb" / "_lamin.db"
+        filepath = self.storage.root / ".lamindb" / "lamin.db"
```
Koncopd, line 301:
```suggestion
        filepath = self.storage.root / ".lamindb/lamin.db"
```

falexwolf: I think what you're suggesting is not going to properly resolve on Windows, is it? That's what the `/` is for

Koncopd: No, it is fine on windows.

falexwolf: Understood that it also works on Windows and you consider it less efficient to use the operator. Let's go with the string!

PR #997, opened 2025-03-19
🚚 Rename `_is_initialized` to `storage_uid.txt`

falexwolf:
Running tests in lamindb here:

- https://github.com/laminlabs/lamindb/pull/2580

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,5 +108,5 @@ lamin_sphinx
 docs/conf.py
 _docs_tmp*
 *.db
-_is_initialized
+storage_uid.txt
 test.ipynb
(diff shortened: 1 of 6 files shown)


falexwolf:
@Koncopd, we're never using the `_is_initialized` file except in the one instance where the legacy filename is checked.

This means that for almost all storage locations, the `_is_initialized` file will _not_ be renamed for the foreseeable future. I don't see why we'd ever want to read this file, so I don't think this is an issue.

Koncopd:
I haven’t looked yet, I will check soon. The main thing where we should care about this is during `delete`.

falexwolf:
> The main place where we should care about this is during delete.

You are correct. If it attempts to delete the legacy file and doesn't find it that'd be bad. But I thought we're just deleting the folder with `rmtree()` after checking that it's empty up to the marker file. If now we're writing the marker file twice because the marking is called more often than we thought, of course, that's also going to be broken. We'll always get an error that the  `.lamindb/` directory isn't yet empty.

PR #999, opened 2025-03-20
🐛 Only call `mark_storage_root` to check write access if the storage record has been just created

Koncopd:
Fixes repetitive writes of `_is_initialized` on `ln.Artifact`  initialization with a cloud storage not in `ln.Storage` table.

`init_storage` is called inside `Artifact` [here](https://github.com/laminlabs/lamindb/blob/01abda450218a9647b1f23b5368193425a0fcf7a/lamindb/models/artifact.py#L180).

### DIFF
--- a/lamindb_setup/core/_settings_storage.py
+++ b/lamindb_setup/core/_settings_storage.py
@@ -162,6 +162,10 @@ def init_storage(
             access_token=access_token,
         )
     # below comes last only if everything else was successful
+    # we have to check hub_record_status here because
+    # _select_storage inside init_storage_hub also populates ssettings._uuid
+    # and we don't want to delete an existing storage record here
+    # only newly created
     if hub_record_status == "hub-record-created":
         try:
             # (federated) credentials for AWS access are provisioned under-the-hood
@@ -172,10 +176,6 @@ def init_storage(
             logger.important(
                 f"due to lack of write access, LaminDB won't manage storage location: {ssettings.root_as_str}"
             )
-            # we have to check hub_record_status here because
-            # _select_storage inside init_storage_hub also populates ssettings._uuid
-            # and we don't want to delete an existing storage record here
-            # only newly created
             if ssettings._uuid is not None:
                 delete_storage_record(ssettings._uuid, access_token=access_token)  # type: ignore
                 ssettings._uuid_ = None
--- a/lamindb_setup/core/_settings_storage.py
+++ b/lamindb_setup/core/_settings_storage.py
@@ -170,7 +170,11 @@ def init_storage(
     # _select_storage inside init_storage_hub also populates ssettings._uuid
     # and we don't want to delete an existing storage record here if no write access
     # only newly created
-    if hub_record_status == "hub-record-created" or not ssettings.type_is_cloud:
+    # local storages not registered in the hub should be also marked
+    is_local_not_retrieved = not (
+        ssettings.type_is_cloud or hub_record_status == "hub-record-retireved"
+    )
+    if hub_record_status == "hub-record-created" or is_local_not_retrieved:
         try:
             # (federated) credentials for AWS access are provisioned under-the-hood
             # discussion: https://laminlabs.slack.com/archives/C04FPE8V01W/p1719260587167489
--- a/lamindb_setup/core/_settings_storage.py
+++ b/lamindb_setup/core/_settings_storage.py
@@ -152,35 +152,35 @@ def init_storage(
         "hub-record-not-created", "hub-record-retireved", "hub-record-created"
     ] = "hub-record-not-created"
     # the below might update the uid with one that's already taken on the hub
-    if not prevent_register_hub:
-        if ssettings.type_is_cloud or register_hub:
-            from ._hub_core import delete_storage_record, init_storage_hub
+    if not prevent_register_hub and (ssettings.type_is_cloud or register_hub):
+        from ._hub_core import delete_storage_record, init_storage_hub
 
-            hub_record_status = init_storage_hub(
-                ssettings,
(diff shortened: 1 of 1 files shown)


lamindb_setup/core/_settings_storage.py
```diff
@@ -152,35 +152,39 @@ def init_storage(
...
-        # discussion: https://laminlabs.slack.com/archives/C04FPE8V01W/p1719260587167489
-        # if access_token was passed in ssettings, it is used here
-        mark_storage_root(ssettings.root, ssettings.uid)  # type: ignore
-    except Exception:
-        logger.important(
-            f"due to lack of write access, LaminDB won't manage storage location: {ssettings.root_as_str}"
+    if not prevent_register_hub and (ssettings.type_is_cloud or register_hub):
+        from ._hub_core import delete_storage_record, init_storage_hub
+
+        # this retireves the storage record if it exists already in the hub
```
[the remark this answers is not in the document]
Koncopd: Thx, fixed

lamindb_setup/core/_settings_storage.py
```diff
@@ -152,35 +152,39 @@ def init_storage(
...
-                auto_populate_instance=not init_instance,
-                created_by=created_by,
-                access_token=access_token,
-            )
-    # below comes last only if everything else was successful
-    try:
-        # (federated) credentials for AWS access are provisioned under-the-hood
-        # discussion: https://laminlabs.slack.com/archives/C04FPE8V01W/p1719260587167489
-        # if access_token was passed in ssettings, it is used here
-        mark_storage_root(ssettings.root, ssettings.uid)  # type: ignore
```
falexwolf, line 170: The fact that this was always called seems like a pretty brutal logical error. I wonder how this made it into the production code. Probably very old.

lamindb_setup/core/_settings_storage.py
```diff
@@ -152,35 +152,39 @@ def init_storage(
...
-            hub_record_status = "hub-record-not-created"
-        ssettings._instance_id = None
+    # below comes last only if everything else was successful
+    # we check the write access here only if the storage record has been just created
+    # or if the storage is local
+    # also we have to check hub_record_status here because
+    # _select_storage inside init_storage_hub also populates ssettings._uuid
+    # and we don't want to delete an existing storage record here if no write access
+    # only newly created
+    if hub_record_status == "hub-record-created" or not ssettings.type_is_cloud:
```
falexwolf, line 173: Can you consolidate the comment above? The previous comment explained why settings._uuid was called. If you can simplify the explanation of why you could drop the `settings._uuid is not None` condition and replace it with `settings.type_is_cloud` that'd be great.

I don't fully understand this mix of old and new explanation and it's important that we make no more mistakes here.

lamindb_setup/core/_settings_storage.py
```diff
@@ -152,35 +152,39 @@ def init_storage(
...
+        try:
+            # (federated) credentials for AWS access are provisioned under-the-hood
+            # discussion: https://laminlabs.slack.com/archives/C04FPE8V01W/p1719260587167489
+            # if access_token was passed in ssettings, it is used here
+            mark_storage_root(ssettings.root, ssettings.uid)  # type: ignore
+        except Exception:
+            logger.important(
+                f"due to lack of write access, LaminDB won't manage storage location: {ssettings.root_as_str}"
+            )
+            if ssettings._uuid is not None:
```
falexwolf, line 183: This is now longer part of the original if condition check, which previously had `and settings._uuid is not None` in it. So it might be that something inconsistent happens now?

Koncopd: This is for example a case of local storage. So i don't think anything inconsistent happens here.

lamindb_setup/core/_settings_storage.py
Koncopd: Yes, this is very old code. It looks pretty bad, i agree, but i don't see why it is brutal as it doesn't break anything actually.

lamindb_setup/core/_settings_storage.py
falexwolf: Some local storage locations are registered in Central. Did you account for this?

lamindb_setup/core/_settings_storage.py
Koncopd: Hm, i don't fully get your question. The previous comment explained why `if hub_record_status == "hub-record-created"` was needed, but in a different place. But the correct place for this check is here, so we actually check write access with marking only if the hub record was created or if the storage is local.

lamindb_setup/core/_settings_storage.py
Koncopd: Yes, not this is fixed.

lamindb_setup/core/_settings_storage.py
```diff
@@ -170,7 +170,11 @@ def init_storage(
     # _select_storage inside init_storage_hub also populates ssettings._uuid
     # and we don't want to delete an existing storage record here if no write access
     # only newly created
-    if hub_record_status == "hub-record-created" or not ssettings.type_is_cloud:
+    # local storages not registered in the hub should be also marked
+    is_local_not_retrieved = not (
+        ssettings.type_is_cloud or hub_record_status == "hub-record-retireved"
```
falexwolf, line 175: There is a typo in the return code! ☺️ -- That has to be fixed in another PR.

Koncopd: Oh, and it is even bigger that than that, but at least it was consistent, so actually nothing got broken due to the typos. https://github.com/laminlabs/lamindb-setup/pull/1001

PR #1012, opened 2025-03-28
🔊 Non-verbose logging of progress

Koncopd:


Koncopd:
I forgot about this, but it should be ready @falexwolf

Koncopd:
Actually, this might be not a good idea at all, as we forward further to the terminal and this suppresses the progress logging.

PR #1015, opened 2025-04-04
🐛 Set db token only on the transaction level

Koncopd:
Paired with:

- https://github.com/laminlabs/lamindb/pull/2641

### DIFF
--- a/lamindb_setup/core/_hub_core.py
+++ b/lamindb_setup/core/_hub_core.py
@@ -466,12 +466,11 @@ def access_db(
     # local is used in tests
     url = f"/access_v2/instances/{instance_id}/db_token"
     if os.environ.get("LAMIN_ENV", "prod") != "local":
-        api_url = instance_api_url
-        if api_url is None:
+        if instance_api_url is None:
             raise RuntimeError(
                 f"Can only get fine-grained access to {instance_slug} if api_url is present."
             )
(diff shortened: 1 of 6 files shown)


falexwolf:
Cool! 😄

lamindb_setup/core/_hub_core.py
```diff
@@ -439,32 +438,45 @@ def _access_aws(*, storage_root: str, client: Client) -> dict[str, dict]:
...
+    instance_slug: str
+    instance_api_url: str | None
+    if isinstance(instance, InstanceSettings):
+        instance_id = instance._id
+        instance_slug = instance.slug
+        instance_api_url = instance._api_url
+    else:
+        instance_id = UUID(instance["id"])
+        instance_slug = instance["owner"] + "/" + instance["name"]
+        instance_api_url = instance["api_url"]
```
falexwolf, line 454: The 9 line preceding lines here are not pretty at all. Why do we have to accept a `dict` here rather than the properly typed `InstanceSettings` object?

Koncopd: It is needed in `lamindb` in `using`. It is better than to initialize `isettings` from the `dict`.

falexwolf (approved):
Looks great!

PR #1016, opened 2025-04-07
🚸 Enable to authenticate and set the current instance through env variables

falexwolf:
Also:

* deprecate `close()` in favor of `disconnect()`
* no longer issue "anonymous" warnings during login

### DIFF
--- a/lamindb_setup/_setup_user.py
+++ b/lamindb_setup/_setup_user.py
@@ -8,7 +8,7 @@ from lamin_utils import logger
 from ._check_setup import _check_instance_setup
 from ._init_instance import register_user
 from .core._settings import settings
-from .core._settings_load import load_or_create_user_settings, load_user_settings
+from .core._settings_load import load_user_settings
 from .core._settings_save import save_user_settings
 from .core._settings_store import (
     current_user_settings_file,
@@ -28,15 +28,12 @@ def load_user(email: str | None = None, handle: str | None = None) -> UserSettin
         save_user_settings(user_settings)  # needed to save to current_user.env
         assert user_settings.email is not None or user_settings.api_key is not None
     else:
-        user_settings = load_or_create_user_settings()
         if email is None:
             raise SystemExit(
                 "✗ Use your email for your first login in a compute environment. "
                 "After that, you can use your handle."
             )
-        user_settings.email = email
-        user_settings.handle = handle
-        save_user_settings(user_settings)
+        user_settings = UserSettings(handle=handle, email=email, uid="null")  # type: ignore
 
     from .core._settings import settings
 
@@ -82,7 +79,7 @@ def login(
         elif user_settings.email is None:
             raise SystemExit(f"✗ No stored user email, please call: lamin login {user}")
     else:
-        user_settings = UserSettings(handle="temporary", uid="00000000")
+        user_settings = UserSettings(handle="temporary", uid="null")
 
     from .core._hub_core import sign_in_hub, sign_in_hub_api_key
 


lamindb_setup/_connect_instance.py
```diff
@@ -219,12 +220,20 @@ def connect(instance: str | None = None, **kwargs) -> str | tuple | None:
     _user: UserSettings | None = kwargs.get("_user", None)
     if _user is not None:
         access_token = _user.access_token
+    if (
+        settings.user.api_key is None
+        and settings.user.password is None
+        and os.environ.get("LAMIN_API_KEY") is not None
+    ):
+        login()
```
Koncopd, line 228: Can we avoid mixing `connect` and `login`?

lamindb_setup/_connect_instance.py
```diff
@@ -219,12 +220,20 @@ def connect(instance: str | None = None, **kwargs) -> str | tuple | None:
...
+    if instance is None:
+        instance = os.environ.get("LAMIN_CURRENT_INSTANCE")
```
Koncopd, line 230:
```suggestion
        instance = os.environ.get("LAMIN_CURRENT_INSTANCE", None)
```
Just for clarity.

lamindb_setup/_connect_instance.py
Koncopd: You should also note that _user might be passed in `kwargs`.

falexwolf: I have a clear use case in mind for this: when a script starts running in a remote environment, e.g., on Modal, the user should be able to set up everything by providing env variables.

It's not possible to create a "python wrapper" that would somehow call `login()` explicitly.

The user code just imports lamindb and should run just like in a local environment.

We could think about creating another user-facing function but I feel that `ln.connect()` in auto-connect mode is perfectly fine for this.

Note that `login()` is only called if the user is not already authenticated _and_ `LAMIN_API_KEY` is set. Do you think this could have any side effects?

Also: `login()` is strictly upstream and orthogonal to `connect()`. So, adding `connect()` to `login()` would be terrible, but the other way round, it's not so terrible, is it?

lamindb_setup/_connect_instance.py
falexwolf: Hm, I had with `None` before but in fact I prefer it without `None` because it's a very well known convention.

lamindb_setup/_connect_instance.py
falexwolf: > You should also note that _user might be passed in kwargs.

Right, this is indeed a problem and there should be a check for this. But first: why is that and why is there no comment explaining it? Or did I just not find that comment?

Koncopd: >But first: why is that and why is there no comment explaining it? Or did I just not find that comment?

Good question, i am trying to recall where it is even used.

Koncopd: Ok, this is mostly a way to pass an access token to connect, i will clarify with a comment.

falexwolf: Let's simply discuss on a call. Maybe you have a better idea!

lamindb_setup/_connect_instance.py
```diff
@@ -301,22 +301,10 @@ def connect(instance: str | None = None, **kwargs) -> str | tuple | None:
...
-        #         ssettings_record.save()
-        #     # raised by django when the access is denied
-        #     except ProgrammingError:
-        #         pass
         load_from_isettings(isettings, user=_user, write_settings=_write_settings)
         if _reload_lamindb:
             importlib.reload(importlib.import_module("lamindb"))
-        else:
-            logger.important(f"connected lamindb: {isettings.slug}")
+        logger.important(f"connected lamindb: {isettings.slug}")
```
Koncopd, line 307: Why removing `else`? It is needed to avoid double message from here https://github.com/laminlabs/lamindb-setup/blob/367fad9ed2f2a64a6b6a5afe67128778b9cb9417/lamindb_setup/_check_setup.py#L162

falexwolf:
Giving it a last check.

PR #1018, opened 2025-04-11
✨ Add `.pt` and `.ckpt` to valid suffixes

Zethson:
`.pt` and `.ckpt` for model weights/checkpoints

### DIFF
--- a/lamindb_setup/core/upath.py
+++ b/lamindb_setup/core/upath.py
@@ -52,7 +52,20 @@ VALID_SIMPLE_SUFFIXES = {
     ".qs",  # https://cran.r-project.org/web/packages/qs/vignettes/vignette.html
     ".rds",
     ".pt",
-    ".ckpt"
+    ".pth",
+    ".ckpt",
+    ".state_dict",
+    ".keras",
+    ".pb",
+    ".pbtxt",
+    ".savedmodel",
+    ".pkl",
+    ".pickle",
+    ".bin",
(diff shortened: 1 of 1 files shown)


falexwolf:
Can we get some vetting from an ML expert or AI for that this is comprehensive? I remember something like `.ptc` 🤔 -- how does hugging face models do it?

And that `.ckpt` is indeed worth supporting?

Zethson:
There's more but this is just a start.

Model file extensions in machine learning vary by framework. Here are the common ones:

- `.pt` / `.pth` - PyTorch
- `.state_dict` - PyTorch state dictionary
- `.keras` - Keras/TensorFlow
- `.pb` / `.pbtxt` / `.savedmodel` - TensorFlow/TensorFlow Lite
- `.pkl` / `.pickle` - Scikit-learn models
- `.bin` / `.safetensors` - Hugging Face Transformers
- `.model` - XGBoost/LightGBM
- `.mlmodel` - Core ML (Apple)
- `.mar` - TorchServe archive

Lightning gives us `.ckpt`.

falexwolf:
Got it! And right  at least `.pth` should also be supported, but I also don't see why not supporting the rest?

Koncopd (approved):

falexwolf:
Given Lukas is now on vacation, @Koncopd, can you add a few more of these suffixes so that we're mostly good on that front?

PR #1020, opened 2025-04-14
✨ Refresh db token on expiration

Koncopd:
This PR allows to refresh expired db tokens (jwt) for instances with fine-grained access.


falexwolf (approved):
Looks great up to missing PR description. Will also ask CoPilot.

Koncopd:
>The enter override references 'token' from the outer closure

Also seems helpful, let me think on it.

PR #1026, opened 2025-04-29
♻️ Add `max_length` to serialized schema

fredericenard:


falexwolf:
This is `max_length` for `CharField`, right? We only need it for validation on the frontend as otherwise the database will throw an error, correct?

Is there anything else we need that we don't yet have?

If we keep adding attributes the JSON will explode at some point. It's inefficiently encoded and I'm a bit worried about the scale of this thing.

The particular problem here is that this option is only relevant for `CharField`, but you're still adding it to any field.

If we encoded this e.g. via:
```
options: {"max_length": 20}.
```
which simply is omitted for fields in which it doesn't apply, then the encoding would be more efficient.

I'm not sure how bad the situation already is with the current JSON.

fredericenard:
> This is `max_length` for `CharField`, right? We only need it for validation on the frontend as otherwise the database will throw an erro

I want to use it to infer the length of non bionty UIDs, see [here](https://github.com/laminlabs/laminhub/issues/2280#issuecomment-2837870560).

fredericenard:
> I'm not sure how bad the situation already is with the current JSON.

It's not too bad actually, there is not a lot of fields but that could be more efficient.

> The particular problem here is that this option is only relevant for CharField, but you're still adding it to any field.

I agree that's not ideal, I modified the implementation so that we only get a `uid_length` defined for the model.

tests/hub-local/test_update_schema_in_hub.py
```diff
@@ -44,6 +44,11 @@ def test_update_schema_in_hub(setup_instance):
     assert not schema["schema_json"]["core"]["artifact"]["is_link_table"]
     assert schema["schema_json"]["core"]["artifactulabel"]["is_link_table"]
 
+    assert schema["schema_json"]["core"]["artifact"]["uid_length"] == 20
+    assert schema["schema_json"]["core"]["transform"]["uid_length"] == 20
+    assert schema["schema_json"]["core"]["bionty"]["gene"] == 20
```
falexwolf, line 49: This test doesn't make much sense, does it? 🤔

tests/hub-local/test_update_schema_in_hub.py
```diff
@@ -44,6 +44,11 @@ def test_update_schema_in_hub(setup_instance):
     assert not schema["schema_json"]["core"]["artifact"]["is_link_table"]
     assert schema["schema_json"]["core"]["artifactulabel"]["is_link_table"]
 
+    assert schema["schema_json"]["core"]["artifact"]["uid_length"] == 20
+    assert schema["schema_json"]["core"]["transform"]["uid_length"] == 20
```
falexwolf, line 48: This should be 16, not 20.

falexwolf:
> I agree that's not ideal, I modified the implementation so that we only get a uid_length defined for the model.

That's way more efficient if we don't need for other CharFields. But say we do want the information in the future for validation purposes, it'd be good to have it, right?

I guess you didn't like the `options: {}` suggestion and I totally see why.

But if in your code (just revert the last commit) you'd omit defining `max_length` for all fields in which it is `None` rather then setting it to be `None`, you'd right away keep the JSON much smaller. Why not do that?

image

fredericenard:
> you'd omit defining max_length for all fields in which it is None rather then setting it to be None, you'd right away keep the JSON much smaller. Why not do that?

The content of the json is exported from the Pydantic model `FieldMetadata` [here](https://github.com/laminlabs/lamindb-setup/blob/0253fd01dbfa4eab39145933698fa0a01b08cd01/lamindb_setup/_schema_metadata.py#L181). 

I can hardcode a rule to exclude it in the for loop. But we should certainly use a different model for some types of fields, there is a lot to improve here.

falexwolf:
We might need to add two more things here.

sunnyosun:
We also need these two attributes to be serialized:
- `registry._ontology_id_field`
- `registry._name_field`

fredericenard:
> We might need to add two more things here.

what?

PR #1041, opened 2025-05-27
✨ Migrate in presence of fine-grained access

falexwolf:
Needs:

- https://github.com/laminlabs/laminhub/pull/3294


Koncopd:
This is fine as a temporary hack but i don't think we want anything like this in `lamindb-setup`.

falexwolf:
Will never be merged.

PR #1046, opened 2025-06-05
✨ Allow to re-connect in the same process

Koncopd:


Koncopd:
@falexwolf here we can test re-connection gin the same process.

PR #1056, opened 2025-06-13
♻️ Do not write instance settings on connect by default

Koncopd:


Koncopd:
`hub-local` fails due to the stale sequence for user after creating an instance via `create_hosted_test_instance` from `laminhub`. It needs to be fixed in `laminhub`.

PR #1059, opened 2025-06-14
♻️ Simplify storage marking & storage-from-hub retrieval

falexwolf:
For:

- https://github.com/laminlabs/lamindb/pull/2841

### DIFF
--- a/lamindb_setup/core/_settings_storage.py
+++ b/lamindb_setup/core/_settings_storage.py
@@ -164,18 +164,8 @@ def init_storage(
             created_by=created_by,
             access_token=access_token,
         )
-    # below comes last only if everything else was successful
-    # we check the write access here only if the storage record has been just created
-    # or if the storage is local
-    # also we have to check hub_record_status here because
-    # _select_storage inside init_storage_hub also populates ssettings._uuid
-    # and we don't want to delete an existing storage record here if no write access
-    # only newly created
-    # local storages not registered in the hub should be also marked
-    is_local_not_retrieved = not (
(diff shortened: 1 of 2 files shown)


lamindb_setup/core/_settings_storage.py
```diff
@@ -164,18 +164,8 @@ def init_storage(
...
-    # or if the storage is local
-    # also we have to check hub_record_status here because
-    # _select_storage inside init_storage_hub also populates ssettings._uuid
-    # and we don't want to delete an existing storage record here if no write access
-    # only newly created
-    # local storages not registered in the hub should be also marked
-    is_local_not_retrieved = not (
-        ssettings.type_is_cloud or hub_record_status == "hub-record-retrieved"
-    )
-    if hub_record_status == "hub-record-created" or is_local_not_retrieved:
```
Koncopd, line 178: This is surely needed in my opinion, please don't merge without a discussion.

falexwolf: What I added is a super set that contains this condition.

falexwolf: We need to broaden the "write check" to the case in which no hub record was created, meaning `"hub-record-not-created"` is returned.

You might not be registering your instance on the hub, but this doesn't invalidate the requirement of the "write check".

falexwolf: One interesting question concerns the case where the instance is not on the hub, but the storage location was already marked before. I believe we should account for this, but it's independent of this change here.

PR #1064, opened 2025-06-16
✨ Add lazy import mechanism to enable model import before django setup

ap--:
This add a lazy proxy mechanism that hooks into pythons import machinery to automatically create lazy class and function wrappers to defer import to when lamin is connected to an instance.

xref: laminlabs/lamindb#2849


falexwolf:
Superseded by:

- https://github.com/laminlabs/lamindb/pull/3580

PR #1092, opened 2025-07-15
✨ Check plugin versions on import

Zethson:
By user/customer request.

- Checks on import whether the versions of all plugins meet a required threshold. Raises an error if not and asks the user to upgrade to at least the required version
- Currently only support bionty, wetlab, & clinicore. Therefore, this would not capture any other plugins.
- I did NOT know which versions we should pin to at the moment so I just went with the latest releases.

### DIFF
--- a/lamindb_setup/__init__.py
+++ b/lamindb_setup/__init__.py
@@ -42,32 +42,35 @@ import importlib.metadata
 import os
 
 from packaging import version as packaging_version
+from packaging.requirements import Requirement
 
-from lamindb_setup.errors import ModuleWasntConfigured
 
-
-def _check_plugin_version(package_name: str, min_version: str) -> None:
+def _check_plugin_compatibility(package_name: str) -> None:
     try:
-        current_version = importlib.metadata.version(package_name)
+        dist = importlib.metadata.distribution(package_name)
+        lamindb_requirement = None
 
-        if packaging_version.parse(current_version) < packaging_version.parse(
-            min_version
-        ):
+        for req_str in dist.requires or []:
+            req = Requirement(req_str)
+            if req.name == "lamindb":
+                lamindb_requirement = req
+                break
+
+        if not lamindb_requirement:
+            return
+
+        if not lamindb_requirement.specifier.contains(__version__):
             raise RuntimeError(
-                f"The version of {package_name} you have ({current_version}) is incompatible "
-                f"with lamindb, please upgrade it: pip install {package_name}>{min_version}"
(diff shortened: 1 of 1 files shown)


falexwolf:
> @Koncopd inconvenient because it has to be maintained? What are the reasons against it?

It's mostly error prone to maintain this in two places.

@ap-- has the solution to this: basing it off of the pyproject.toml

And yes, we had similar logic in lamindb a long time ago but I forgot what the actual downside was/ why we removed it. 🤔 -- We had it in Scanpy since the very early days in 2017 and I think similar checks are still there and actually helped avoid user frustration.

So, I think I'm in favor for as long as this isn't hardcoded.

ap--:
> clinicore is currently not pinned

I think that it doesn't have a lower bound is a bug. newest lamindb doesn't work with an older clinicore.

falexwolf:
> I disagree, because the missing lower bound in the the clinicore extra in lamindb is the root cause for the lamindb update issue that dan reported.

I don't fully agree.

I think that it's _very likely_ that the lower bounding in a lamindb extra would *not* have prevented what Dan experienced because he, like many others, likely forgot to pass the `clinicore` extra and ran:

```shell
pip install -U lamindb
pip install -U 'lamindb[bionty,jupyter]'  #  from docs
```

If he had run

```shell
pip install -U 'lamindb[bionty,jupyter,clinicore]'
```

he would *not* have run into the problem.

Hence, while your example is formally correct I don't think that it's likely that somebody would purposefully pin an outdated version of clinicore against a recent release of lamindb:
```
pip install "lamindb[clinicore]==1.8.0" "clinicore==1.1.0"
```

I think the actual problem is the missing upper bound in clinicore. We have one that says `lamindb<2` but it would have had to be `lamindb<1.6`. That's the same unfortunate situation that we have in bionty & wetlab.

IMO, the presence of extras with bounds in lamindb provides some convenience but actually violates DRY for coding the bounding. Adding a lower bound for the extra for wetlab & bionty might have prevented a hiccup, and for consistency clinicore _should_ also have had it. But much better would have been to carefully architect the bounds in the packages themselves, just like we do for any python package that's used together with other python packages.

ap--:
Hi Lamins,

Now that since lamindb">=1.9.0" clinicore has a lower bound in lamindb's pyproject.toml, this be could redone with the logic inverted, couldn't it? It would at least then cover the bionty, clinicore and wetlab errors that users run into when updating only lamindb without extras.

falexwolf:
I don't think we've had issues with this topic anymore due to `lamindb` pinning the lightweight official public modules in the default install and `lamindb-core` presenting the minimal install.

PR #1121, opened 2025-08-15
🐛 Fix issues with loading migrations and unavailable instances

falexwolf:


lamindb_setup/_migrate.py
```diff
@@ -85,7 +85,6 @@ def create(cls) -> None:
         setup_django(settings.instance, create_migrations=True)
 
     @classmethod
-    @disable_auto_connect
```
falexwolf, line 88: I don't know why we had this here. With the new setup it understandably breaks `lamin migrate deploy`. I don't think we need it so I'm removing it.

lamindb_setup/_connect_instance.py
```diff
@@ -304,7 +304,10 @@ def connect(instance: str | None = None, **kwargs: Any) -> str | tuple | None:
                 else:
                     from lamindb_setup.core.django import reset_django
 
-                    if settings.instance.slug != "none/none":
+                    if (
+                        settings._instance_exists
```
falexwolf, line 308: `settings.instance` is not available if `settings._instance_exists` is `False`, hence we need to add this check.

PR #1131, opened 2025-09-05
♻️ Refactor and prettify login function

falexwolf:


lamindb_setup/_setup_user.py
```diff
@@ -43,15 +43,35 @@ def load_user(email: str | None = None, handle: str | None = None) -> UserSettin
...
+        api_key: User API key. The CLI `lamin login` command prompts if `api_key` is `None`.
     """
+    if user is None:
+        if api_key is None:
+            if "LAMIN_API_KEY" in os.environ:
+                api_key = os.environ["LAMIN_API_KEY"]
+            else:
+                api_key = input("Your API key: ")
+    else:
+        api_key = None
```
Koncopd, line 61: I don't think this is correct, it should raise an error, otherwise confusing in my opinion, it is not immediately clear for a user if it uses `user` or `api_key` if both provided.

Koncopd: overall the logic should be restructured because further checks below like `if user is None and api_key is None` don't make much sense anymore i think.

falexwolf: I see that just below these sanity checks are already happening. There was some duplication across the CLI and the Python API implementations. That's why I moved the logic out of lamin-cli.

image

I'll clean things up.

Koncopd: I have made some edits, feel free to revert if you don't agree.

Koncopd: sorry, i tried to push mine but it was rejected due to your commit.

falexwolf: Ohhhh, I mean i didn't spend much thought in my commits; can you take a look and add what you think is missing?

Koncopd: there are still some strange things in your recent commit, can i force push my fixes?

falexwolf:
Perfect, thanks Sergei!

PR #1143, opened 2025-09-23
🗃️ Do not assume `account.id` is always equal to `account.user_id` referencing `auth.user.id`

Koncopd:
It is assumed now that in our supabase table `account` `id` is always equal to `user_id` (referencing `auth.user.id`). We might need to break this in the future due to some SSO related migrations. This ensures that the potential change won't affect our login flow.

### DIFF
--- a/lamindb_setup/core/_hub_core.py
+++ b/lamindb_setup/core/_hub_core.py
@@ -661,7 +661,12 @@ def _sign_in_hub(email: str, password: str, handle: str | None, client: Client):
             "password": password,
         }
     )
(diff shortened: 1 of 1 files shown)


Koncopd:
Ok, this is correct but in `get-jwt-v1` for issuing JWTs from api keys account id is used instead of user id, i will change that.
https://github.com/laminlabs/laminhub/blob/bdbea6d64d19ac420b0d76ad29adc887242e6146/rest-hub/supabase/functions/get-jwt-v1/index.ts#L59

PR #1154, opened 2025-10-06
🐛 Fix small chunksize for very large files when uploading to s3

Koncopd:
Fix this error for very large files:
image


falexwolf:
Great!

Issue #1160, opened 2025-10-09
Implement `load_clone`


Zethson:
Was not a part of the scope of https://github.com/laminlabs/lamindb-setup/pull/1149

PR #1183, opened 2025-10-28
✨ Add clone utilities

Zethson:
Fixes https://github.com/laminlabs/lamindb-setup/issues/1160

- Adds `init_local_sqlite` to initialize local clones from existing postgres instances. This is basically a one step solution for both `io.export_db` and `io.import_db`.
- Adds `connect_local_sqlite` and `connect_remote_sqlite` which connect to local and remote clones respectively. Note that these change the current connection.
- Adds `upload_sqlite_clone` which allows stores clones on the remote storage.

### DIFF
--- a/lamindb_setup/core/_clone.py
+++ b/lamindb_setup/core/_clone.py
@@ -134,40 +134,6 @@ def connect_remote_sqlite(instance: str, *, copy_suffix: str | None = None) -> N
     isettings._is_clone = True
     isettings._persist(write_to_disk=True)
 
-    # TODO This used to be the full download code that also enabled reading compressed files
-    """
-    cloud_db_path = str(ln_setup.settings.instance.storage.root) + "/.lamindb/lamin.db"
-    sqlite_file_path_gz = create_path(cloud_db_path + ".gz")
-    sqlite_file_path = create_path(cloud_db_path)
-
-    local_sqlite_target_path = (
-        ln_setup.settings.cache_dir
-        / _strip_cloud_prefix(ln_setup.settings.instance.storage.root_as_str)
-        / ".lamindb"
-        / "lamin.db"
-    )
-
-    if not local_sqlite_target_path.exists() or overwrite:
-        local_sqlite_target_path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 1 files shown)


lamindb_setup/core/_clone.py
```diff
@@ -90,4 +102,116 @@ def connect_local_sqlite(instance: str) -> None:
 
     isettings = load_instance_settings(settings_file)
     isettings._persist(write_to_disk=False)
-    isettings._load_db()
+    isettings._load_db(sqlite_read_only=read_only)
```
Koncopd: the locking log appers due to this in `_load_db` https://github.com/laminlabs/lamindb-setup/blob/881a219a1a99d21bb22efa17f9198067168212c7/lamindb_setup/core/_settings_instance.py#L617

lamindb_setup/core/_clone.py
```diff
@@ -91,3 +104,115 @@ def connect_local_sqlite(instance: str) -> None:
     isettings = load_instance_settings(settings_file)
     isettings._persist(write_to_disk=False)
     isettings._load_db()
```
Koncopd: just don't use it here. Just use

```
from lamindb_setup._check_setup import disable_auto_connect

from .django import setup_django

disable_auto_connect(setup_django)(self)
```

lamindb_setup/core/_clone.py
```diff
@@ -91,3 +104,115 @@ def connect_local_sqlite(instance: str) -> None:
...
+    Args:
+        instance: Instance slug in the form `account/name` (e.g., `laminlabs/privatedata-local`).
+        copy_suffix: Optional suffix of the local clone.
+        overwrite: Whether to overwrite existing local database.
+    """
+    import lamindb_setup as ln_setup
+
+    # Step 1: Create the settings file
+    # We need to connect to the real instance to get the settings
+    ln_setup.connect(instance)
```
Koncopd: do this instead https://github.com/laminlabs/lamindb-setup/blob/881a219a1a99d21bb22efa17f9198067168212c7/lamindb_setup/_connect_instance.py#L100

PR #1191, opened 2025-11-09
♻️ Include implicit link tables in schema

fredericenard:


lamindb_setup/_schema_metadata.py
```diff
@@ -418,14 +418,10 @@ def _get_modules_metadata(self):
...
+        for model in apps.get_models(include_auto_created=True):
             # Check if model meets the criteria
-            if (
-                model.__class__ is Registry
-                and model is not SQLRecord
-                and not model._meta.abstract
-            ):
-                module_name = model.__get_module_name__()
+            if model is not SQLRecord and not model._meta.abstract:
+                module_name = Registry.__get_module_name__(model)
```
fredericenard: @falexwolf let me know if you see a better way to get the module name.

fredericenard:
> Yes, maybe we should first check if this unblock Robrecht.

Related thread: https://laminlabs.slack.com/archives/C08J590666Q/p1762874557638859?thread_ts=1762529577.319259&cid=C08J590666Q

PR #1192, opened 2025-11-10
⚡ Speedup IO of DB import/export

Zethson:
This PR greatly speed ups the export & import of databases. Whereas before it took several hours for big databases, we can now do so in a few seconds.

**Export (`export_db`):**

1. **Parallel processing**
   - Old: Exported tables one at a time
   - New: Uses `ProcessPoolExecutor` with configurable `max_workers`

2. **PostgreSQL uses COPY TO**
   - Old: `pd.read_sql_table()` for everything
   - New: `cursor.copy_expert()` for Postgres. We're now streaming directly from CSV to leverage good ol' optimized C code & have bulk optimized operations for postgres.

3. **Large SQLite tables are chunked**
   - Old: Loaded entire table into memory
   - New: Reads in chunks if >500K rows, merges at the end
   - Added `chunk_size` parameter

**Import (`import_db`):**

1. **JSON serialization is vectorized**
   - Old: Applied lambda to every row with `.apply()`
   - New: Uses mask to find rows needing serialization, then `.map()` only on those
   - Added numpy array support

2. **PostgreSQL uses COPY FROM**
   - Old: `df.to_sql()` for everything
   - New: Converts df to CSV buffer, uses `cursor.copy_expert()` for Postgres

3. **SQLite gets multi-row inserts**
   - Old: Single-row inserts via plain `df.to_sql()`
   - New: Added `method='multi'` with dynamic chunking (stays under SQLite's 999 variable limit)

4. **SQLite performance PRAGMAs**
   - Temporarily disables `synchronous`, uses `journal_mode = MEMORY`, increases cache
   - All restored in finally block

5. **Bug fixes**
   - Fixed duplicate dialect check
   - Added empty dataframe handling
   - Proper cleanup with try/finally

### DIFF
--- a/lamindb_setup/io.py
+++ b/lamindb_setup/io.py
@@ -51,7 +51,22 @@ def _export_full_table(
     directory: Path,
     chunk_size: int,
 ) -> list[tuple[str, Path]] | str:
-    """Export a full table to parquet."""
+    """Export a registry table to parquet.
+
+    For PostgreSQL, uses COPY TO which streams the table directly to CSV format,
+    bypassing query planner overhead and row-by-row conversion (10-50x faster than SELECT).
+
+    For SQLite with large tables, reads in chunks to avoid memory issues when tables exceed available RAM.
+
+    Args:
+        registry_info: Tuple of (module_name, model_name, field_name) where field_name
+            is None for regular tables or the field name for M2M link tables.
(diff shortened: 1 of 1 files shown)


falexwolf (approved):
Great! 😄

PR #1231, opened 2025-12-08
🏗️ Re-connect django rather than reset django

falexwolf:
One alternative design for us to study, @Koncopd.


falexwolf:
Was implemented here:

- https://github.com/laminlabs/lamindb/pull/3580

PR #1334, opened 2026-05-01
🔥 Remove `UPathStr` type

Koncopd:
Deprecated


falexwolf:
This would be a breaking change if somebody used this type in their third party package, right?

We should first deprecate before removing?

PR #1340, opened 2026-05-08
✨ Enable `ln.connect()` in marimo notebooks

ishitajain9717:
Adding patch for allowing tracking in marimo notebooks

### DIFF
--- a/lamindb_setup/core/django.py
+++ b/lamindb_setup/core/django.py
@@ -22,6 +22,12 @@ IS_SETUP = False
 IS_MIGRATING = False
 CONN_MAX_AGE = 299
 
+def _is_running_in_marimo() -> bool:
+    try:
+        import marimo
+        return bool(marimo.running_in_notebook())
+    except Exception:
+        return False
 
 def get_connection(connection_name: str):
     from django.db import connections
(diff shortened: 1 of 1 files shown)


lamindb_setup/core/django.py
```diff
@@ -23,6 +23,15 @@
 CONN_MAX_AGE = 299
 
 
+def _is_running_in_marimo() -> bool:
```
falexwolf, line 29: This is great, @ishitajain9717! Super clean and surgical.

One thing to check: We should compare the runtime of this check with a check that would first check for whether marimo is installed rather than trying to import it? Maybe AI has a clear opinion on this and we don't need to check.

I'm asking because there are many things that happen upon import of lamindb and we need to be very cautious about keeping this as performant as can be.

It might be that triggering an `ImportError` is extermely (milli-second) fast. In that case this is here is the correct solution.

If it's not and `"marimo" in sys.modules` or similar is faster than we should do what is faster.

ishitajain9717: I check and I made the changes, if marimo isn’t already loaded, we return False immediately and never import it. That keeps the check cheap and matches how marimo works, the kernel imports marimo before user cells run.

lamindb_setup/core/django.py
```diff
@@ -319,7 +336,7 @@ def setup_django(
     view_schema: bool = False,
     appname_number: tuple[str, int] | None = None,
 ):
-    if IS_RUN_FROM_IPYTHON:
+    if IS_RUN_FROM_IPYTHON or _is_running_in_marimo():
```
Koncopd, line 339: maybe better to cache it like like with `IS_RUN_FROM_IPYTHON`

falexwolf: If we need it elsewhere: agreed! If we only need in django setup no need for extra complexity. ☺️ 

We'll see what's necessary in `lamindb`.

